Issue #10, opened 2021-05-26
[MXNET] Activation in yaml and calibration algorithm


PawelGlomski-Intel:
Hi, I would like to contribute and improve MXNet integration and add MXNet 1.9 support, but one thing is not clear to me.

I would like to ask what exactly does the "activation" of an operator mean in "capabilities", especially its "algorithm" attribute. I thought this meant the operator's input (as stated [here](https://github.com/intel/lpot/blob/master/docs/adaptor.md#query-api)), and the "algorithm" attribute determines whether minmax or kl will be used when calibrating that operator's input tensors, but I believe this is not what is happening in MXNet adaptor.
https://github.com/intel/lpot/blob/v1.3.1/lpot/adaptor/mxnet.py#L581-L592
Here, `calib_minmax_layers ` and `calib_kl_layers`  are filled with the names of the operators' outputs, not their inputs. Is this intentional?

Thanks

ftian1:
Pawel, thanks for your contribution.

as for how to explain the acivation field of query API, we treat it as input and output datatype at the same time. so fp32 means fallback to FP32. u8 means u8 in u8 out. s8 means s8 in s8 out.

PR #11, opened 2021-05-27
Improve MXNet integration

PawelGlomski-Intel:
### Summary
Improving MXNet integration. This change will focus on improving the current features.

### Motivation
- mxnet.yaml file is filled with configurations, which do not affect quantization but increase tuning space
- quantizable_ops are not selected properly (which also increase tuning space)
- calib_<kl/minmax>_layers are not selected properly (when quantizable_ops selection is fixed, this is unrelated to #10)
- if confirmed, #10 must be fixed

### Changes
In mxnet.yaml:
- [x] using only "int8" for operators that are meant to be quantized (and discard uint8) until it is possible in MXNet to select dtype of quantization per node/op (now "auto" dtype is hardcoded in MXNet adaptor)
- [x] removed granularity until it is possible in MXNet to select the granularity of quantization per node/op (now "tensor-wise" is hardcoded in MXNet adaptor)
- [x] removed bf16 until it is supported in LPOT for MXNet 
- [x] fixed names of operators

In mxnet.py:
- [ ] quantizable_ops selection
- [ ] calib_<kl/minmax>_layers selection

Please review the structure of my 1.9.0 configuration and decide whether this can be applied to the rest of the yaml file.


lpot/adaptor/mxnet.yaml
```diff
@@ -125,6 +125,71 @@
...
+
+  ops:
+    int8: ['_sg_mkldnn_conv', 'Convolution', '_sg_mkldnn_fully_connected', 'FullyConnected', '_sg_mkldnn_selfatt_qk', '_sg_mkldnn_selfatt_valatt']
+    fp32: ['*'] # '*' means all op types
+
+  capabilities:
+    'int8': {
+        '_sg_mkldnn_conv': {
+            'activation': {
+                'dtype': ['int8', 'fp32'],
```
PawelGlomski-Intel, line 143: Would it be possible to add "autoint8" to the possible values of "dtype" attribute, which would correspond to the automatic selection of dtype according to the calibration result of each tensor?
```suggestion
                'dtype': ['autoint8', 'fp32'],
```
This way we could describe more clearly what is actually happening in MXNet.

PR #20, opened 2021-07-22
Update onnxrt_mid.py

peiwenhuang27:
Fixed multiple inputs of ONNX session in the same manner as in `onnxrt.py`


ftian1:
thanks for contributing this PR. I will close it at first as it in fact has been fixed in our internal

PR #179, opened 2022-11-28
Add optimization level for tuning strategy

yiliu30:
## Type of Change

- Feature
 - API changed or not: Add `optimization_level` to select the tuning strategy

## Description

detail description 
JIRA ticket: [ILITV-2358](https://jira.devtools.intel.com/browse/ILITV-2358)

- [x] Add conservative traverse logic
      - Start with fp32 model, convert op(s) into lower-precision(`bf16`/`fp16`/`int8`).
      - Convert ops into lower precision by op-type wise.
      - Convert op into lower precision one by one based on the result of the previous step.
- [x] Add optimization level
      - `0`: using conservative strategy
      - `1`: using basic strategy by default
- [x] UTs 
- [ ] Doc, will be submitted by another PR 

- Pre-CI 
- Extension test [JOB3599](https://inteltf-jenk.sh.intel.com/job/intel-lpot-validation-top-mr-extension/3599/artifact/report.html)

## Dependency Change?

None


neural_compressor/strategy/conservative.py
```diff
@@ -122,7 +122,7 @@ def traverse(self):
             self.objectives.baseline = self.baseline
             # self.best_tune_result = self.baseline
             # Initialize the best qmodel as fp32 model
-            self.best_qmodel = self._fp32_model
+            # self.best_qmodel = self._fp32_model
```
ftian1, line 125: why this line gets commented off?

yiliu30: We will initialize the `self.best_qmodel` with the first tuning result whose accuracy meets the requirements. [code](https://github.com/intel/neural-compressor/blob/7d46e5ad87680035207eedf10fc0126a9313f86d/neural_compressor/strategy/conservative.py#L159)
If we initialize the `self.best_qmodel` with the fp32 model, we may return the fp32 model as the tuning result which may confuse the user, right?

PR #231, opened 2022-12-07
Enhance ITEX benchmark UT check

lvliang-intel:
Signed-off-by: Lv, Liang1 <liang1.lv@intel.com>

## Type of Change

bug fix
API not changed

## Description

detail description 
JIRA ticket: ILITV-2480

ITEX benchmark UT check will not has issue when unexcepted log file exist.

UT test.

## Dependency Change?

No.

### DIFF
--- a/test/itex/test_tensorflow_itex_basic.py
+++ b/test/itex/test_tensorflow_itex_basic.py
@@ -73,6 +73,12 @@ class TestItexEnabling(unittest.TestCase):
               accuracy:
                 metric:
                   topk: 1
+              performance:
+                warmup: 10
+                iteration: 100
+                configs:
+                  cores_per_instance: 4
+                  num_of_instance: 7
             tuning:
                 strategy:
                   name: basic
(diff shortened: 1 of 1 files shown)


test/itex/test_tensorflow_itex_basic.py
```diff
@@ -258,8 +258,11 @@ def test_itex_benchmark_gpu(self):
             evaluator('performance')
 
         found_multi_instance_log = False
+        num_of_instance = int(os.environ.get('NUM_OF_INSTANCE'))
```
ftian1, line 261: we should remove all environment variable usage from our code.

is this the only way to get num_of_instance and core_per_instance?

test/itex/test_tensorflow_itex_basic.py
```diff
@@ -254,12 +262,15 @@ def test_itex_benchmark_gpu(self):
 
             evaluator = Benchmark('fake_yaml_2.yaml')
             evaluator.b_dataloader = common.DataLoader(dataset)
+            num_of_instance = evaluator.conf.usr_cfg.evaluation.performance.configs.num_of_instance
```
ftian1, line 265: sorry, I don't understand why this UT needs to be updated? it doesn't work before?

at INC 2.0 we are focusing on new API. I suppose such old UT shouldn't be updated, why such change is needed?

lvliang-intel: This UT will check the log file name. Sometimes it will fail if the test folder has unexpected '.log' file. This fix is used to avoid such failure.

chensuyue (approved):

PR #243, opened 2022-12-08
add keras-in/keras-out to INC

ClarkChin08:
## Type of Change

Feature with new keras adaptor supported

## Description
  1. support keras-level transformation instead of freezing to graph_def
  2. support specific Q/DQ/QDense/QConv2D op in keras level
  3. support mnist model to keras

UT and keras model should be added

## Dependency Change?

No


examples/keras/mnist/mnist.py
```diff
@@ -0,0 +1,105 @@
...
+    return score[1]
+
+def main():
+    x_train, y_train, x_test, y_test = build_dataset()
+    model = build_model(x_train, y_train, x_test, y_test)
+
+    from neural_compressor.quantization import fit
+    from neural_compressor.config import PostTrainingQuantConfig
+    from neural_compressor.utils.utility import set_random_seed
+    from neural_compressor.experimental import common
```
ftian1, line 95: should not to use experimental in new API example. PR244 is used to provide new dataset/dataloader interface to use. you can refine this part when having time.

lvliang-intel: yes, will refine it after PR244 merged.

PR #278, opened 2022-12-15
Add recipe for TRT EP

mengniwang95:
Signed-off-by: Mengni Wang <mengni.wang@intel.com>

## Type of Change

task

## Description

TRT EP has some limitations, add recipe for it to enable QDQ models with TRT EP.


hshen14 (approved):
LGTM

PR #284, opened 2022-12-15
Reconstructing pruning API derived from Component.py.

YIYANGCAI:
Signed-off-by: YIYANGCAI <yiyang.cai@intel.com>

## Type of Change

API changed or not

## Description

detail description 
JIRA ticket: xxx

## Dependency Change?


neural_compressor/conf/pythonic_config.py
```diff
@@ -18,7 +18,7 @@
 import logging
 from .dotdict import DotDict
 from ..config import _BaseQuantizationConfig, accuracy_criterion, BenchmarkConfig, \
-                     check_value, DistillationConfig, options, PruningConfig
+                     check_value, DistillationConfig, options, WeightPruningConfig
```
ftian1, line 21: here should we introduce new WeightPruningConfig and keep old PruningConfig as the latter is used by 1.x pruning implementation?

ftian1: so pls check if INC 1.14.2 includes PruningConfig or not. if yes, it means 1.x pruning code used this definition before and should keep and add new WeightPruningConfig but not replace?

wenhuach21: As you mentioned before, PruningConfig should be only used in 2.0a. I have checked  neural_compressor/config.py and nerual_compressor/conf/config in v1.4.2, this class has not been defined

wenhuach21: > here should we introduce new WeightPruningConfig and keep old PruningConfig as the latter is used by 1.x pruning implementation?
As you mentioned before, PruningConfig should be only used in 2.0a. I have checked neural_compressor/config.py and nerual_compressor/conf/config in v1.4.2, this class has not been defined

examples/pytorch/nlp/huggingface_models/question-answering/pruning/pytorch_pruner/eager/run_qa_no_trainer.py
```diff
@@ -57,8 +56,8 @@
 from transformers.utils import check_min_version
 from transformers.utils.versions import require_version
 from utils_qa import postprocess_qa_predictions
-from neural_compressor.pruning import Pruning
-from neural_compressor.pruner.utils import WeightPruningConfig
+from neural_compressor.pruner.pruning import Pruning
```
ftian1, line 59: Pruning under pruner is not correct here. let's have quick sync

ftian1 (approved):

PR #289, opened 2022-12-16
Remove Docs of Model Conversion Class

zehao-intel:
Signed-off-by: zehao-intel <zehao.huang@intel.com>

## Type of Change

API changed or not: Yes

## Description

detail description 
JIRA ticket: NA

the expected behavior that triggered by this PR: as title

how to reproduce the test (including hardware information): preci

## Dependency Change?

any library dependency introduced or removed: No


ftian1:
sorry, maybe misleading you about this. We don't need ModelConversion in 2.0 API but needs to keep it in experimental/ dir for backward compatibility.

so we shouldn't make any change in Experiemental dir.

PR #425, opened 2023-01-05
Fix online document build issue

NeoZhangJianyu:
## Type of Change

bug fix for online document

## Description

1. correct the versions.html path in all html files.
2. fix git operation error issue.

## Dependency Change?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 <div align="center">
   
-Intel® Neural Compressor1
+Intel® Neural Compressor
 ===========================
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -1,6 +1,6 @@
 <div align="center">
   
-Intel® Neural Compressor
+Intel® Neural Compressor1
```
ftian1, line 3: does it make sense to have such number? can we have another naming?

NeoZhangJianyu: Sorry, it's changed to for test. Now it's restored.

WenjiaoYue (approved):

PR #458, opened 2023-01-12
fix bug of export scale info

xin3he:
Signed-off-by: Xin He <xin3.he@intel.com>

## Type of Change
Fix bug of export scale info, before the scale info is collected from last sample.


PenghuiCheng (approved):

ftian1:
preci can discover such bug or shall we add new UT for that?

xin3he:
> preci can discover such bug or shall we add new UT for that?

UT cannot find it. we should enable an accuracy check in example. I will do it in next spirit.

PR #500, opened 2023-02-01
Add model inplace flag for quantization

PenghuiCheng:
## Type of Change

Add model inplace flag for quantization  
API changed

## Description

Add model inplace flag for quantization  

model will be inplace in quantization in neural compressor.

Local tested


neural_compressor/conf/config.py
```diff
@@ -789,7 +789,8 @@ def percent_to_float(data):
         'framework': And(str, lambda s: s in list(FRAMEWORKS.keys()) + ['NA']),
         Optional('inputs', default=[]): And(Or(str, list), Use(input_to_list)),
         Optional('outputs', default=[]): And(Or(str, list), Use(input_to_list)),
- 
+        Optional('model_inplace', default=False): And(bool, lambda s: s in [True, False]),
```
hshen14, line 792: Can we put under quantization/? framework is general and is not supposed to have quantization specific implementation.

PenghuiCheng: It is under model field， not framework.

PenghuiCheng:
duplicated to https://github.com/intel/neural-compressor/pull/541

PR #514, opened 2023-02-03
Add new config items and support smooth quant

mengniwang95:
## Type of Change

feature

## Description

detail description 
JIRA ticket: https://jira.devtools.intel.com/browse/ILITV-2675


mengniwang95:
domain detection, tuning alpha of smooth_quant, tuning recipes will be added in another PR, please review API for smooth_quant and config first

mengniwang95:
@ftian1

neural_compressor/conf/pythonic_config.py
```diff
@@ -28,6 +28,8 @@ def __init__(self,
                  inputs=[],
                  outputs=[],
                  backend='default',
+                 domain='auto',
```
ftian1, line 31: why changing here? this should be legacy conf code, right?

neural_compressor/config.py
```diff
@@ -348,6 +348,8 @@ def __init__(self,
                  inputs=[],
                  outputs=[],
                  backend="default",
+                 domain="auto",
```
ftian1, line 351: will you add function comments here to explain them?

ftian1 (approved):
the code is ok for me. pls add some function comments to make it readable, especially the func comments

neural_compressor/conf/pythonic_config.py
mengniwang95: > 
it maps new config items to the legacy conf, so the legacy conf should have these attributes. Keep these attrs but don't take them as initialization params is ok?

ftian1 (approved):

PR #628, opened 2023-03-09
Fix mix precision

mengniwang95:
## Type of Change

bug fix

## Description

Fix mix precision


mengniwang95:
> please also fix in https://github.com/intel/neural-compressor/blob/mengni/mix/docs/source/quantization_mixed_precision.md#turn-off-auto-mixed-precision-during-quantization

I will add ORT part and @changwangss @lvliang-intel please check PT/TF parts

ftian1 (approved):

PR #631, opened 2023-03-09
Quantization/sq example update

wenhuach21:
## Type of Change
example change


wenhuach21:
benchmarking has weird behavior, please help correct it.

chensuyue (approved):

PR #632, opened 2023-03-09
Enable automatic model slim for channel pruning and MHA pruning.

YIYANGCAI:
Signed-off-by: Cai, Yiyang <yiyang.cai@intel.com>

## Type of Change

new feature
API not changed

## Description

An enhancement of channel pruning. In some particular structure we can remove the all-zero channels and thus the model's inference speed can be significantly improved. HF folks recently mentioned similar feature in pruning therefore we would like to enable such feature in INC quickly.

Automatically detect target pattern in the model and slim it by removing some all-zero channels

Pre-CI and example tests.

## Dependency Change?

None.


neural_compressor/compression/pruner/model_slim/pattern_analyzer.py
```diff
@@ -0,0 +1,381 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import torch
+import re
+
+JIT_SUPPORT_OPS = ['linear', 'gelu', 'mul'] # linear and all act_fn supported by pytorch-aten extension
```
wenhuach21, line 22: too strict constraint, need to improve in the next version

neural_compressor/compression/pruner/model_slim/pattern_analyzer.py
```diff
@@ -0,0 +1,381 @@
...
+                    else:
+                        continue 
+        # step 2: dfs  
+        # execute dfs-based pattern matching
+        dfs()
+        print(f"Matched {len(self.search_results)} pattern {self.target_pattern} in model {type(self.model).__name__}")
+        JitBasicSearcher.get_layer_for_all(self)
+        return self.search_results
+
+class Linear2LinearSearcher(JitBasicSearcher):
```
wenhuach21, line 319: could we implement it with some pattern rather than a class

wenhuach21 (approved):
significant improvement is needed in the next version

neural_compressor/compression/pruner/model_slim/pattern_analyzer.py
YIYANGCAI: Will do! All improvements including functions and coding structure will be carried out in next version.

YIYANGCAI: MHA slim will also be supported, since hf seems to have similar features.

neural_compressor/compression/pruner/model_slim/weight_slim.py
```diff
@@ -0,0 +1,319 @@
...
+        ffn_indice = PostCompressionUtils.get_mha_input_indice(self.ffn.weight, self.hidden_size, self.head_nums)
+        all_indice_to_prune = {
+            "query": query_indice,
+            "key": key_indice,
+            "value": value_indice,
+            "ffn": ffn_indice
+        }
+        
+        # alignment, take the least heads to prune
+        print(all_indice_to_prune)
```
wenhuach21, line 305: ->log

PR #687, opened 2023-03-16
pruning/add block mask feature

WeiweiZhang1:
## Type of Change

feature

API not changed


neural_compressor/compression/pruner/patterns.py
```diff
@@ -19,11 +19,10 @@
 from .utils import torch
 from .utils import logger
 from collections import namedtuple
-
+import torch.nn.functional as F
```
wenhuach21, line 22: use lazy import, otherwise ut will fail

examples/pytorch/nlp/huggingface_models/question-answering/pruning/eager/run_qa_no_trainer_block.py
wenhuach21, line 1: please also help run one of the snip-moment examples to check whether there are any regressions

PR #718, opened 2023-03-21
Enabled multi-nodes pruning of ResNet50 on cpu device.

XinyuYe-Intel:
## Type of Change

Example enabling
API not changed

## Description

Enabled multi-nodes pruning of ResNet50 on cpu device.

results on multi-nodes is comparable with single node.

follow readme

## Dependency Change?

no library dependency introduced or removed


chensuyue:
Please update spell check dict.

chensuyue (approved):

PR #739, opened 2023-03-24
Neural Insights initial design

bmyrcha:
## Type of Change

Added Neural Insights module for diagnosis and profiling
Modified Quantization and Benchmark configs by adding respectively `diagnosis` and `profiling` parameters.

## Description

Created Neural Insights backend code and modifications to Quantization and Benchmark API.

Neural Insights workloads added when executing Quantization API with `diagnosis` and Benchmark API with `profiling` option. These workloads should be visible in web GUI. Backend also should provide detailed information about specified diagnosis or profiling via endpoints.

Limited manual tests using TF mobilenet_v1 example.

## Dependency Change?

No

### DIFF
--- a/neural_insights/ux/bin/neural_insights
+++ b/neural_insights/ux/bin/neural_insights
@@ -1,0 +1,24 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+# Copyright (c) 2023 Intel Corporation
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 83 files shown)


neural_insights/ux/bin/neural_insights
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env python3
```
chensuyue, line 1: do we need to wrap up by `ux`

PR #757, opened 2023-03-28
ExportConfig&Torch2ONNXConfig docstring polish

xin3he:
## Type of Change

add docstring

## Description

ExportConfig&Torch2ONNXConfig docstring polish

N/A

N/A

## Dependency Change?

N/A

### DIFF
--- a/neural_compressor/config.py
+++ b/neural_compressor/config.py
@@ -1246,7 +1246,19 @@ class MixedPrecisionConfig(PostTrainingQuantConfig):
             self._precision = precision
 
 class ExportConfig:
-    """Config Class for Export."""
+    """Common Base Config for Export.
+
+    Args:
+        dtype (str, optional): The data type of the exported model, select from ["fp32", "int8"]. 
+                               Defaults to "int8".
+        opset_version (int, optional): The ONNX opset version used for export. Defaults to 14.
+        quant_format (str, optional): The quantization format of the exported int8 onnx model, 
+                                      select from ["QDQ", "QLinear"]. Defaults to "QDQ".
+        example_inputs (optional): Example inputs used for tracing model. Defaults to None.
+        input_names (list, optional): A list of model input names. Defaults to None.
+        output_names (list, optional): A list of model output names. Defaults to None.
+        dynamic_axes (dict, optional): A dictionary of dynamic axes information. Defaults to None.
+    """
     def __init__(
         self,
         dtype="int8",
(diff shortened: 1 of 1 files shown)


neural_compressor/config.py
```diff
@@ -1343,7 +1355,37 @@ def __init__(self):
...
+        quant_format (str, optional): The quantization format of the exported int8 onnx model, 
+                                      select from ["QDQ", "QLinear"]. Defaults to "QDQ".
+        example_inputs (required): Example inputs used for tracing model. Defaults to None.
+        input_names (list, optional): A list of model input names. Defaults to None.
+        output_names (list, optional): A list of model output names. Defaults to None.
+        dynamic_axes (dict, optional): A dictionary of dynamic axes information. Defaults to None.
+        recipe (str, optional): A string to select recipes used for Linear -> Matmul + Add, select from 
+                                ["QDQ_OP_FP32_BIAS", "QDQ_OP_INT32_BIAS", "QDQ_OP_FP32_BIAS_QDQ"]. 
+                                Defaults to 'QDQ_OP_FP32_BIAS'.
+
```
mengniwang95, line 1373: I think it is better if you can provide a link to different recipes so that users can understand what does they mean

neural_compressor/config.py
```diff
@@ -1343,7 +1355,37 @@ def __init__(self):
...
-    """Config Class for Torch2ONNX."""
+    """Config Class for Torch2ONNX.
+
+    Args:
+        dtype (str, optional): The data type of the exported model, select from ["fp32", "int8"]. 
+                               Defaults to "int8".
+        opset_version (int, optional): The ONNX opset version used for export. Defaults to 14.
+        quant_format (str, optional): The quantization format of the exported int8 onnx model, 
+                                      select from ["QDQ", "QLinear"]. Defaults to "QDQ".
+        example_inputs (required): Example inputs used for tracing model. Defaults to None.
```
mengniwang95, line 1366: type for example_inputs?

xin3he: it has lots of kinds, like dict, tuple, list, tensor, I cannot specify it.

neural_compressor/config.py
xin3he: @chensuyue , Do we need to add a link to docs in docstring?

chensuyue: I think we don't need a link in docstring.

neural_compressor/config.py
chensuyue: We should list the types that we accept.

PR #795, opened 2023-04-10
Add doc for developer about how to add new data type

yiliu30:
## Type of Change

documentation 

## Description
How to add a new data type
[The rendered version](https://github.com/intel/neural-compressor/blob/ly/new_data_doc/docs/source/add_new_data_type.md)

## Dependency Change?

None


docs/source/add_new_data_type.md
```diff
@@ -0,0 +1,91 @@
...
+conf = PostTrainingQuantConfig(op_type_dict=op_type_dict)
+...
+
+```
+
+With this code, all `Conv2d` operators will be quantized to 4-bit, with weight using `int4` and activation using `uint4`.
+
+
+### Others
+It's important to note that INC currently only supports N-bit quantization for the PyTorch backend, where N is an integer between 1 and 7. While other backends or quantization schemes may be added in the future.
```
ftian1, line 91: what we should mention is the flexible design of INC to easily extend to other new data type.

ftian1: and pls add a link to `advance` page for user reference. 

The title of this doc should be something like "How to support new data type, like int4, with few line changes". 

the whole idea should highlight our extensible design like I commented at above.

docs/source/add_new_data_type.md
```diff
@@ -0,0 +1,154 @@
...
+```
+1. **Strategy**: Drives the overall tuning process and utilizes `adaptor.query_fw_capability` to query the framework's capabilities.
+
+2. **Adaptor**: Parses the framework YAML, filters some corner cases, and constructs the framework capability. This includes the capabilities of each operator and other model-related information.
+
+3. **Strategy**: Constructs the tuning space based on the framework capability and initiates the tuning process.
+
+4. **Strategy**: Generates the tuning configurations for each operators of the model using the tuning space constructed in the previous step, specifying the desired tuning process.
+
+5. **Adaptor** Invokes the specific kernels for the calibration and quantization based on the tuning configuration.
```
ftian1, line 40: lack of a separator ":"

ftian1 (changes requested):
others looks good to me.

PR #814, opened 2023-04-18
Fix bug for slim non-bias linear layers.

YIYANGCAI:
## Type of Change

bugfix
API changed or not: None

## Description

Fix a bug in auto slim when processing linear layers whose bias attribute is None.

Able to slim non-bias linear layers

PreCI

## Dependency Change?

None

### DIFF
--- a/neural_compressor/compression/pruner/model_slim/pattern_analyzer.py
+++ b/neural_compressor/compression/pruner/model_slim/pattern_analyzer.py
@@ -168,7 +168,7 @@ class JitBasicSearcher(object):
             self.placeholder_shape = [1, 16]
             self.placeholder_dtype = torch.int64
         else:
-            logger.warning("Cannot generate dummy input automatically, please set it manually when initialzation.")
+            logger.warning("Cannot generate dummy input automatically.")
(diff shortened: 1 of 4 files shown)


WeiweiZhang1 (approved):

neural_compressor/compression/pruner/model_slim/pattern_analyzer.py
```diff
@@ -183,44 +183,56 @@ def generate_static_graph_with_dummyinput(self):
                 torch.ones([1, 80, 3000]),
                 torch.ones([1, 448], dtype=torch.int64)
```
wenhuach21, line 184: it is quite tricky , is there any chance to polish it

wenhuach21 (approved):

neural_compressor/compression/pruner/model_slim/pattern_analyzer.py
YIYANGCAI: This will not be called if we provide an external datasets for whisper model. However, it is still hard to analyze whisper's dummy input.

PR #830, opened 2023-04-25
Smooth Quant for Tensorflow backend

Spycsh:
## Type of Change

Feature

## Description

We support Conv2d and MatMul smoothquant for Tensorflow/ITEX backend

SmoothQuant is used to decrease the accuracy loss for CV and NLP models

Invoke UT cases
distilbert_base
opt-125m
gpt2-medium

## Dependency Change?

None


neural_compressor/adaptor/tf_utils/smooth_quant_scaler.py
```diff
@@ -0,0 +1,148 @@
...
+                    cur_const_node = W_const_node_lst[w_i]
+                    try:
+                        scale = np.power(A_max_per_in_channel, self.alpha) /  \
+                                np.power(W_max_per_in_channel, (1-self.alpha))
+                    except ValueError as e:
+                        logger.info(e)
+                        logger.info("Skip smoothing the node: {}".format(cur_const_node.name))
+                        continue
+                    # clip the scales that are too small
+                    scale = np.clip(scale, a_min=1e-2, a_max=1e8)
```
wenhuach21, line 138: may I know whether there are some models that benefited from this setting? We may apply this setting to other backends.

wenhuach21:
Also, it seems that group convolution is not excluded, which will throw an exception when applying sq. Depthwise conv is ok

neural_compressor/adaptor/tf_utils/smooth_quant_scaler.py
Spycsh: Hi, I'm not sure which line of code you are refering to that makes some "benefit"?

If you mean the try-catch clock from line 130-136, we catch the exception because I remember that in the resnet there is a const node that will cause "operands could not be broadcast together with shapes...", which means that the two inputs' shape do not match for the computation. Also at that time we check that intel tensorflow does not support optimization on that specific setting so we just skip that.

If you mean the line 138 clipping of the scales, I have no idea on that line of code :) because I read about PyTorch and ONNX backend sq code and find this line. I think it may skip the "divide by zero" problem (but for numpy I think it only will raise a Warning rather than an Error), so I do not really think this line make sense.

If you mean the line 139-142, this few lines (need uncommented) is useful for the distilbert. I use that because I think smoothquant is not really a silver bullet for restoring accuracy. I have done some experiments to see that there are some cases when the MeanSquareError increases after using smoothquant, compared to not using sq (we expect MSE to be lower after using sq). The paper assumes that 1) the activation >> weight for some channels 2) for the same channel, variance is relatively small. However, I think this is too ideal for arbitrary models. For example, I find that for one distilbert model I have to uncomment 139-142 to get the accuracy correct. That is just my experience and it may not be correct, but it may works if you meet relevant problems for other backends.

Spycsh:
> Also, it seems that group convolution is not excluded, which will throw an exception when applying sq. Depthwise conv is ok

Yes, it seems that what you mentioned is related to the try-catch clock from line 130-136. The backend supported patterns may be inadequate here. We will recheck it later.

neural_compressor/adaptor/tf_utils/smooth_quant_scaler.py
Spycsh: Thanks for pointing out. I remember that min value does not make some clear effect or benefit. Maybe I just forgot to edit it back. I will check and fix that.

examples/tensorflow/large_language_model/README.md
```diff
@@ -0,0 +1,24 @@
+1. Get fp32 accuracy
```
chensuyue, line 1: The example structure is like `examples/[framework]/[model domain]/[model name]/[optimization method]/...`, so please refactor the example path for LLM models.

chensuyue:
Please update the example path and unify the example interface, let's do some extension test for the new add models.

examples/tensorflow/large_language_model/README.md
Spycsh: OK, got it

Spycsh:
Hi @chensuyue , ITEX backend is now supported. do we need to add another entry parameter that specify the backend?

examples/tensorflow/nlp/large_language_models/quantization/ptq/smoothquant/requirements.txt
```diff
@@ -0,0 +1,2 @@
+tensorflow
+datasets
```
chensuyue, line 2: Test require `transformers`, https://inteltf-jenk.sh.intel.com/job/intel-lpot-validation-top-mr-extension/5021/artifact/linux/tensorflow/gpt2_medium_sq/tensorflow-gpt2_medium_sq-linux-spr-tune.log

PR #833, opened 2023-04-26
sq autotuning/algorithm enhance

wenhuach21:
## Type of Change

enhancement and bugfix


chensuyue:
Does this PR target v2.2?

PR #840, opened 2023-04-27
Remove some redundant classes for quantization, benchmark and mixed precision

PenghuiCheng:
## Type of Change

Code quality  
API changed

## Description

Remove some redundant classes for quantization, benchmark and mixed precision

quantization/mixed precision/benchmark work fine. 

Local tested


neural_compressor/config.py
```diff
@@ -1667,7 +1667,7 @@ class MixedPrecisionConfig(object):
     def __init__(self,
                  device="cpu",
                  backend="default",
-                 precision="bf16",
+                 precisions="bf16",
```
chensuyue, line 1670: Please update the readme

chensuyue:
Can we unify the parameter `config` and `conf` in `def fit(..., config=None, ...)`?   
For `neural_compressor/benchmark.py`, `neural_compressor/mix_precision.py`, `neural_compressor/quantization.py`.

xin3he (approved):

docs/source/mixed_precision.md
```diff
@@ -42,8 +42,8 @@ Supported precisions for mix precision include bf16 and fp16. If users want to g
 from neural_compressor import mix_precision
 from neural_compressor.config import MixedPrecisionConfig
 
-conf = MixedPrecisionConfig(precision='bf16')
-converted_model = mix_precision.fit(model, config=conf)
+conf = MixedPrecisionConfig(precisions='bf16')
```
ftian1, line 45: precisions='bf16' has been the default value. we can remove this for simplicity

neural_compressor/benchmark.py
```diff
@@ -135,391 +135,257 @@ def get_bounded_threads(core_ids, threads, sockets):
...
+            import intel_extension_for_pytorch
+        assert sys.platform in ['linux', 'win32'], 'only support platform windows and linux...'
+        # disable multi-instance for running bechmark on GPU device
+        set_all_env_var(conf)
+
+    config_instance(raw_cmd)
+    summary_benchmark()
+
+
+def fit(model, conf, b_dataloader=None, b_func=None):
```
ftian1, line 383: is it incompatible change as previous parameter is config, but now it's conf.

and it's inconsistent between those user facing APIs, like Quantization.fit & MixedPrecision.fit. some of them are Optional, but some of them are not

neural_compressor/compression/__init__.py
```diff
@@ -16,6 +16,4 @@
 # limitations under the License.
 
 from .callbacks import QuantizationAwareTrainingCallbacks, DistillationCallbacks, PruningCallbacks
-from ..experimental.compression import prepare_pruning
-from .. import WeightPruningConfig
 from .pruner.model_slim.auto_slim import model_slim, parse_auto_slim_config
```
ftian1, line 19: this import looks not necessary?

neural_compressor/quantization.py
```diff
@@ -21,372 +21,14 @@
 import random
 import numpy as np
 from .config import _Config, options
-from .model.model import BaseModel, get_model_fwk_name, get_model_type, Model, MODELS
+from .data.dataloaders.dataloader import check_dataloader
```
ftian1, line 24: suggest to export those symbols at the top namespace.

for example, 

from .data import check_dataloader
from .metric import register_customer_metric

ftian1 (changes requested):
most of them are ok for me. I leave some comments here, pls revise them

neural_compressor/benchmark.py
PenghuiCheng: now, all fit functions use conf, because quantization used conf, so we align it to conf.

ftian1 (approved):

ftian1:
> Can we unify the parameter `config` and `conf` in `def fit(..., config=None, ...)`? For `neural_compressor/benchmark.py`, `neural_compressor/mix_precision.py`, `neural_compressor/quantization.py`.

I have same concerns here. if we unify to conf, is it a incompatible change comparing with INC 2.1 release?

PR #846, opened 2023-04-28
fix metric in mixed precision

violetch24:
## Type of Change

fix metric to match current changes in mixed precision
API not changed 

preci & extension test


chensuyue:
Merge into #840

PR #851, opened 2023-05-03
fix document, example readme and code

xin3he:
## Type of Change

bug fix

## Description

Readme is not correct to fit new design and enhance example capability to support more models.

Existing test pass.

local test on other kind of model in torchvision, such as densenet121.

### DIFF
--- a/docs/source/smooth_quant.md
+++ b/docs/source/smooth_quant.md
@@ -309,13 +309,13 @@ conv2d/linear->conv2d/linear/layernorm/batchnorm/instancenorm/t5norm/llamanorm/g
 ```
 
 ## Validated Models
-neural_compressor: 2.1
+Neural Compressor: 2.1
 
 IPEX: 2.0
 
 Dataset: lambada
 
-task: text-generation
+Task: text-generation
 
 alpha [0.4, 0.6] is sweet spot region in SmoothQuant paper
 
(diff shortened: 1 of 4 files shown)


violetch24 (approved):

docs/source/smooth_quant.md
```diff
@@ -351,9 +351,9 @@ smooth_quant_args description:
 
 "alpha": "auto" or a float value. Default is 0.5. "auto" means automatic tuning.
 
-"folding":
-- False: Allow inserting mul to update the input distribution and not absorbing. IPEX can fuse inserted mul automatically and folding=False is recommended. And for PyTorch FBGEMM backend, folding=False setting will only convert model to QDQ model.
-- True: Only allow inserting mul with the input scale that can be absorbed into the last layer. 
+"folding": whether to fold torch.mul into the last layer.
```
hshen14, line 354: it's not only for pytorch. we may remove torch

docs/source/smooth_quant.md
```diff
@@ -351,9 +351,9 @@ smooth_quant_args description:
...
+- False: Allow inserting mul to update the input distribution and no folding. IPEX (version>=2.1) can fuse inserted mul automatically and folding=False is recommended. For Stock PyTorch, setting folding=False will only convert the model to a QDQ model.
```
hshen14, line 355: How about below?
False: Stock PyTorch and Intel Extension for PyTorch (version < 2.1)
True: Intel Extension for PyTorch (version >= 2.1) and ONNX Runtime?

docs/source/smooth_quant.md
xin3he: Yes, I agree.

docs/source/smooth_quant.md
hshen14: 1. It's framework backend capability. so the better way is to define such config in backend yaml. User should not care about folding at front-end unless we want to override backend capability. Shall we do that?
2. If we agree with 1, we don't need to worry about 2 in user config. but we may need to give a support matrix for SmoothQuant across frameworks and backends in SmoothQuant readme. Please talk with Mengni.

xin3he: Okay

hshen14 (approved):
LGTM

PR #900, opened 2023-05-22
Add weight attribute to SQLinearWrapper class

changwangss:
## Type of Change

to enable t5 series smooth quant.

## Description
to avoid inference failed. https://github.com/huggingface/transformers/blob/main/src/transformers/models/t5/modeling_t5.py#L318
```
-> isinstance(self.wo.weight, torch.Tensor)
(Pdb) self.wo.weight
*** AttributeError: 'SQLinearWrapper' object has no attribute 'weight'
(Pdb) self.wo
SQLinearWrapper(
  (sq_linear): Linear(in_features=2816, out_features=1024, bias=False)
)
```

## Dependency Change?

### DIFF
--- a/neural_compressor/adaptor/pytorch.py
+++ b/neural_compressor/adaptor/pytorch.py
@@ -1351,10 +1351,12 @@ class TemplateAdaptor(Adaptor):
             def __init__(self, module):
                 super().__init__()
                 self.add_module('sq_linear', module)
-                self.weight = self.sq_linear.weight
 
             def forward(self, X):
                 return self.sq_linear(X)
+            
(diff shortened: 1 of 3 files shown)


neural_compressor/adaptor/pytorch.py
```diff
@@ -1354,6 +1354,9 @@ def __init__(self, module):
 
             def forward(self, X):
                 return self.sq_linear(X)
+            
+            def weight(self):
```
xin3he, line 1359: miss property

xin3he (approved):

PR #913, opened 2023-05-25
Update Neural Insights setup.py and docs

bmyrcha:
## Type of Change

Documentation

API not changed

## Description

Modified way of providing `neural_insights` executable, included GUI files to setup.py and added initial documentation.

Neural Insights should be successfully installed via command `python setup.py install neural_insights`.

Tested by installing via `python setup.py install neural_insights` and executing `neural_insights` after instalation.

## Dependency Change?

None


neural_insights/README.md
```diff
@@ -0,0 +1,110 @@
+<div align="center">
+  
+Intel® Neural Insights
```
chensuyue, line 3: We could not use `Intel@`, this name is not formal approved.

setup.py
```diff
@@ -65,7 +75,7 @@ def fetch_requirements(path):
         include_package_data=True,
         package_data=package_data,
         install_requires=install_requires_list,
-        scripts=scripts_list,
+        entry_points=entry_points,
```
chensuyue, line 78: what's the benefit of this change?

bmyrcha: It is a recommended approach for cross-platform compatibility:
https://packaging.python.org/en/latest/guides/distributing-packages-using-setuptools/#scripts

neural_insights/README.md
bmyrcha: Done

chensuyue:
Some file is empty `neural_insights/docs/source/installation_guide.md`

PR #944, opened 2023-06-05
Allow dictionary model inputs for onnx export

echarlaix:
In this PR we enable the possibility to give a dictionary as model inputs for the ONNX export, allowing the ONNX export of model such as GPT2, for which an ordered tuple is not enough : needs to explicitely specify the input names as so that the `attention_mask` are not considered as [`past_key_values`](https://github.com/huggingface/transformers/blob/v4.29.2/src/transformers/models/gpt2/modeling_gpt2.py#L1051)


xin3he (approved):

chensuyue:
Azurepipeline run Code-Scan

chensuyue (approved):

PR #948, opened 2023-06-07
fix smooth quant calibration issue

changwangss:
## Type of Change

for opt and some models, when make smoothquant calibration with ipex, the inference only 1 time success due to the default histogramobserver.
will offline test llama, gptj, bloom.

## Description

detail description

## Dependency Change?


changwangss:
close the PR, @xin3he and @changwangss  will continue trace opt models workaround for ipex latest version.

PR #961, opened 2023-06-09
Fixed trace issue when input is dictionary

maktukmak:
## Type of Change

Bug fix

## Description

When Smoothquant is on and folding is True, model tracing for obtaining quantizable ops does not work when the input is a dictionary.

## Expected Behavior & Potential Risk

Smoothquant will work when the input to the model is dictionary.

PR #954, opened 2023-06-08
output check added after sq transform

maktukmak:
## Type of Change

A mathematical check for smoothquant transform

## Description

This check ensures the output of the full model stays the same after smoothquant transformation. This check can catch issues when there are skip connections from layernorms, i.e., layer norm output goes to multiple ops. When folding is enabled, these skip connections create incorrect results. I will create a separate PR to fix that issue.

## Expected Behavior & Potential Risk

None. It is just a check.

PR #962, opened 2023-06-09
Fix for layernorm skip connections in sq

maktukmak:
## Type of Change

Bug fix

## Description

When folding is enabled in Smoothquant, layer norms are used as absorbing layers for linear layers. However, when there are skip connections from the layer_norms, i.e., the output of layer norm goes to multiple ops, smoothquant quant transform produces wrong results. 

An example graph is shown below:

image

which occurs in the BridgeTower model. This PR checks if any layer norm ops have multiple children, in which case, it is not used as absorbing layer.

Smoothquant transform won't produce incorrect results when an absorb layer norm has skip connections

## Dependency Change?

This PR works best after applying the following PRs: #961 and #954

### DIFF
--- a/neural_compressor/adaptor/torch_utils/smooth_quant.py
+++ b/neural_compressor/adaptor/torch_utils/smooth_quant.py
@@ -858,7 +858,8 @@ class GraphTrace:
                     # Check if parent has multiple children of unsupported layers
                     set_inp_type = set(dict_inp[parent])
                     set_inp_type.discard('aten::size')
-                    if (set_inp_type.intersection(self.could_absorb_layers) == set_inp_type) or (set_inp_type.intersection(self.skip_ops_to_find_absorb) == set_inp_type):
+                    if ((set_inp_type.intersection(self.could_absorb_layers) == set_inp_type) or
+                       (set_inp_type.intersection(self.skip_ops_to_find_absorb) == set_inp_type)):
                         prev_absorb_layer.append(parent)
                     else:
                         prev_absorb_layer.append(None)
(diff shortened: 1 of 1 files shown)


xin3he (approved):

chensuyue:
Please fix DCO and Pylint scan.

wenhuach21 (approved):

neural_compressor/adaptor/torch_utils/smooth_quant.py
```diff
@@ -855,7 +855,14 @@ def get_prev_absorb_layer(self, nodes):
                     parent = get_parent(parent)
                     continue
                 if parent.kind() in self.could_absorb_layers:
-                    prev_absorb_layer.append(parent)
+                    # Check if parent has multiple children of unsupported layers
+                    set_inp_type = set(dict_inp[parent])
```
wenhuach21, line 859: the code has some issues. We should check output nodes of  the parent

wenhuach21 (changes requested):

neural_compressor/adaptor/torch_utils/smooth_quant.py
```diff
@@ -875,10 +882,17 @@ def get_absorb_to_layer(self, model, example_input, op_types):
         traced_model = self.trace(model, example_input)
         if traced_model == None:
             return None, None
+        
+        dict_inp = defaultdict(list)
+        for node in traced_model.graph.nodes():
+            inp_list = list(node.inputs())
```
wenhuach21, line 888: we should check the node's output

wenhuach21 (changes requested):

yintong-lu (commented):
Hi, I think there's a mistake when you are creating dict_inp. The input-types of all nodes are saved rather than the output-type.

chensuyue:
Hi @maktukmak, since there are some issues in this PR, we submit a new PR to fix this bug, https://github.com/intel/neural-compressor/pull/1000. So I am going to close this one. Thanks for your submission!

PR #1001, opened 2023-06-15
Copy the model for smoothquant in quantization

PenghuiCheng:
## Type of Change

Copy the model for smoothquant in quantization  
No API changed

## Description

The smoothquant method needs a deep copy of the model in the process of getting quantizable ops, so we also add a deep copy in the quantization process.

Quantize flan-t5 successfully with smoothquant. 

Local tested

### DIFF
--- a/neural_compressor/adaptor/pytorch.py
+++ b/neural_compressor/adaptor/pytorch.py
@@ -3267,14 +3267,11 @@ class PyTorch_IPEXAdaptor(TemplateAdaptor):
         assert not self.version.release < Version("2.1").release, \
             "IPEX version >= 2.1 is required for SmoothQuant."
 
-        if not self.performance_only:
-            try:
-                self.tmp_model = copy.deepcopy(model)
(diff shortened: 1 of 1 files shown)


xin3he (approved):

changwangss (approved):

wenhuach21:
May I know why we need a deepcopy?  If a deepcopy is a must, could we delete the tmp model immediately after using.

xin3he:
> May I know why we need a deepcopy? If a deepcopy is a must, could we delete the tmp model immediately after using.

It's for encoder-decoder model, we need keep the original decoder model when quantizing the decoder model with past kv. Simply, Users can do it themselves in script. 
As we have a deepcopy when getting capability, this deepcopy won't exceed the peak of memory foot print.
@ftian1 , If we can change the order of getting capability and smoothQuant, I will move this deepcopy to the example script.

chensuyue:
Do we need this for v2.2?

PR #1008, opened 2023-06-16
remove recover when sq check failed

xin3he:
## Type of Change

bug fix

## Description

opt-125m requires large atol when checking output, to avoid this error again, we remove recover func and use log instead.

UT pass

local test

## Dependency Change?

N/A


neural_compressor/adaptor/torch_utils/smooth_quant.py
```diff
@@ -780,8 +780,8 @@ def output_is_equal(self, out1, out2, atol=1e-05):
                 return torch.all(torch.isclose(out1, out2, atol=atol))
             return False
         except:
-            logger.warning("Automatically check failed, \
-                            Please check equal manually between out_pre_sq and out_post_sq if necessary.")
+            logger.warning("Automatically check failed, Please check equal manually "
```
wenhuach21, line 783: equal->equivelancy ?

wenhuach21 (approved):

PR #1009, opened 2023-06-16
PT supports weight only

wenhuach21:
## Type of Change

support weight only quantization
## Description

detail description

## Dependency Change?


neural_compressor/adaptor/torch_utils/weight_only.py
```diff
@@ -0,0 +1,146 @@
...
+            weight = m.weight
+        q_weight = quant_weight(weight, num_bits, group_size, schema)
+        if m.__class__.__name__ == "Conv2d":
+            q_weight = q_weight.reshape(orig_shape[1], orig_shape[0], orig_shape[2], orig_shape[3])
+            q_weight = q_weight.permute(1, 0, 2, 3)
+        m.weight.data.copy_(q_weight)
+    return model
+
+#
+# class QuantizerW:
```
hshen14, line 102: remove comments

neural_compressor/adaptor/pytorch.py
```diff
@@ -4434,6 +4457,192 @@ def calculate_op_sensitivity(self, model, dataloader, tune_cfg, output_op_names,
...
+    def GPTQ_quantize(self, model, tune_cfg, dataloader):
+        if 'gptq_args' in self.recipes:
+            percdamp = self.recipes['gptq_args'].get('percdamp', 0.01)
+        # GPTQ(model, dataloader, w_bit, group_size, percdamp=0.01)
+        # TODO: implementation
+        return model
+
+    def AWQ_quantize(self, model, tune_cfg, dataloader):
+        # set default value if has args in recipes, else we use function 
+        if 'gptq_args' in self.recipes:
```
xin3he, line 4558: This is a wrong args. To avoid double pre-ci, I will fix it after this PR is merged.

neural_compressor/config.py
```diff
@@ -28,24 +28,31 @@
...
+    "weight_only": "post_training_weight_only",
 }
 
 
 ops_schema = Schema({
     Optional('weight', default=None): {
         Optional('granularity'): And(
             list,
-            lambda s: all(i in ['per_channel', 'per_tensor'] for i in s)),
+            lambda s: all(i in ['per_channel', 'per_tensor', 'per_block'] for i in s)),
```
hshen14, line 39: shall we align the term with paper recommended, per_group?

neural_compressor/strategy/utils/constant.py
```diff
@@ -22,14 +22,16 @@
 LOWER_BIT_LIST = ['int4']
 
 TUNING_ITEMS_LST = [('activation','scheme'), ('activation','algorithm'), ('activation','granularity'),
-                    ('weight','scheme'), ('weight','algorithm'), ('weight','granularity'), 'sampling_size']
+                    ('weight','scheme'), ('weight','algorithm'), ('weight','granularity'),
+                    ('weight','bit'), ('weight','group_size'), 'sampling_size']
```
hshen14, line 26: here we used group_size so per-block should be renamed to per-group

neural_compressor/adaptor/torch_utils/weight_only.py
xin3he: ok

neural_compressor/config.py
xin3he: OK, I agree.

neural_compressor/adaptor/pytorch.py
```diff
@@ -4434,6 +4457,192 @@ def calculate_op_sensitivity(self, model, dataloader, tune_cfg, output_op_names,
...
+
+        # For tensorboard display
+        self.tune_cfg = tune_cfg
+        self.tune_cfg["approach"] = self.approach
+        self.tune_cfg["framework"] = "pytorch"
+        assert self.approach=='post_training_weight_only', "Please make sure the approach is weight_only"
+
+        q_model._model = self.RTN_quantize(q_model._model, tune_cfg)
+        q_model._model = self.GPTQ_quantize(q_model._model, tune_cfg, dataloader)
+        q_model._model = self.AWQ_quantize(q_model._model, tune_cfg, dataloader)
```
hshen14, line 4515: why trying three algos?

hshen14 (commented):
PR looks good. Can we also add a documentation to show the initial support of weight-only 4-bits at least RTN?

BTW, does it support weight-only 8-bits? I assume YES. Then we should test and mention in the document as well.

neural_compressor/adaptor/pytorch.py
xin3he: This is the design to combine different algos, the quantization is controlled by the tune_cfg.
For example, if there is no GPTQ in tune_cfg, we will return the input model.
The order can be changed with the best recipe we have, I will add logs in each quantize function to show the order for users.

test/quantization/test_weight_only_quantization.py
```diff
@@ -0,0 +1,40 @@
...
+        pass
+
+    def test_conv(self):
+        quant_model_weight_only(self.model, num_bits=3, group_size=-1)
+    #
+    # def test_conv_memory_format(self):
+    #     import  copy
+    #     model = copy.deepcopy(self.model)
+    #     model = model.to(memory_format=torch.channels_last)
+    #     quant_model_weight_only(self.model, num_bits=3, group_size=-1)
```
yiliu30, line 40: Missed the test entry?
```
if __name__ == "__main__":
    unittest.main()
```

xin3he:
> PR looks good. Can we also add a documentation to show the initial support of weight-only 4-bits at least RTN?
> 
> BTW, does it support weight-only 8-bits? I assume YES. Then we should test and mention in the document as well.

Yes, we support 8 bit, the document will be added.

neural_compressor/strategy/strategy.py
```diff
@@ -1260,6 +1269,8 @@ def _set_framework_info(self, q_dataloader, q_func=None):
                     {"default_qconfig": self.config.op_name_dict['default_qconfig']})
             framework_specific_info.update({"q_func": q_func})
             framework_specific_info.update({"example_inputs": self.config.example_inputs})
+            if self.config.approach =='post_training_weight_only':
+                framework = 'pytorchweightonly'   # use specific adaptor for weight_only approach
```
yiliu30, line 1273:
```suggestion
                framework = 'pytorch_weight_only'   # use specific adaptor for weight_only approach

```

xin3he:
> PR looks good. Can we also add a documentation to show the initial support of weight-only 4-bits at least RTN?
> 
> BTW, does it support weight-only 8-bits? I assume YES. Then we should test and mention in the document as well.

Yes, we support 8 bit, the document will be added.

neural_compressor/adaptor/pytorch.py
hshen14: If user specifies multiple algos, then we can do it by internal order. However, I don't see any logics to check what algos should be used.

test/quantization/test_weight_only_quantization.py
xin3he: Thanks, will fix it.

xin3he:
BTW, I removed the granularity and use group_size=-1 to indicate per-channel.

hshen14 (approved):
LGTM

neural_compressor/adaptor/pytorch.py
xin3he: It's WIP, and will be inside the self.gptq_quantize function.

PenghuiCheng (approved):

terryk66:
anaconda prompt

PR #1027, opened 2023-06-20
add deepcopy for SQ folding=True

changwangss:
## Type of Change

bug fix
root cause: when `folding=True` and `eval_func` doesn't provide, the SQ input is ipex prepared model, it used `aten::alias` wrap the op and cause the `linear` and `conv` not to be recognized.
improve: add the deepcopy for this condition, because `folding=False` also need `deepcopy`, so remove the ipex version limitation.

remaining:
there is a remaining curious issue in UT `python test_smooth_quant.py TestSqLinearOpFuse.test_sq_quant_ipex`, for actual models(such as bloom, llama), `tmp_model` and `model` should use the same address. But the previous ut pass, the reason shows `tmp_model=model` create 2 address for `tmp_model` and `model` when get_quantizable ops, I think may be due to the small size, please root cause @xin3he 

## Description

detail description

## Dependency Change?

### DIFF
--- a/neural_compressor/adaptor/pytorch.py
+++ b/neural_compressor/adaptor/pytorch.py
@@ -3064,7 +3064,7 @@ class PyTorch_IPEXAdaptor(TemplateAdaptor):
                     logger.warning("Smoothquant for ipex requires a deepcopy of model"
                                     + ", please avoid out of memory.")
                     try:
-                        tmp_model = copy.deepcopy(model)
+                        tmp_model = copy.deepcopy(model) # Deepcopy due to model changed when `ipex.quantization.prepare`
                     except Exception as e:  # pragma: no cover
                         logger.warning("Fail to deep copy the model due to {}, inplace is used now.".format(
                             repr(e)))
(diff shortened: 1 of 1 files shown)


xin3he (approved):

hshen14:
Why adding deepcopy back? @xin3he did you sort out with IPEX team to see whether one-time calibration/quantization is enough?

xin3he:
> Why adding deepcopy back? @xin3he did you sort out with IPEX team to see whether one-time calibration/quantization is enough?

Yeah, we will raise JIRA to them.

hshen14 (approved):
LGTM

xin3he:
The root cause of this issue is that when we use `inplace=True` in `ipex.prepare`, both model will get hooks and cause trace error while having different ids
In UT, the prepared model got hooks while the other one is pure.
I have raised this unexpected behavior to IPEX team internal. [internal link](https://jira.devtools.intel.com/browse/IPB-1767)

PR #1029, opened 2023-06-21
Supported add hook for root module in intermediate layers KD.

XinyuYe-Intel:
## Type of Change

feature 
API not changed

## Description

Supported add hook for root module in intermediate layers KD.

support add hook for root module in intermediate layers KD

pass UT

## Dependency Change?

no library dependency introduced or removed


hshen14:
better to add a UT to cover this

XinyuYe-Intel:
> better to add a UT to cover this

added.

PR #1036, opened 2023-06-25
pruning/refactor

wenhuach21:
Changes：

1 decouple torch/tf   DONE
2 change mask from float to bool   DONE
3 change score  from float to fp16/bf16   DONE
4 change score/mask from per-weight to per-pattern   DONE
5 re-introduce purning.py  WIP in other PR
6 polish the code   DONE
7 split patten to multi files   DONE
8 split pruning to multi files   WIP in other PR

## Type of Change

    feature
    API not changed

   UTs and examples


neural_compressor/compression/pruner/criteria.py
```diff
@@ -181,7 +182,10 @@ def __init__(self, modules, config):
         assert self.config.end_step > 0, "please set end_step > 0 for gradient based criterion"
         for key in modules.keys():
             p = modules[key].weight
-            self.scores[key] = torch.zeros(p.shape).to(p.device)
+            dtype = torch.float32
+            if self.low_memory_usage:
+                dtype = torch.bfloat16 if p.device.type == 'cpu' else torch.float16
```
wenhuach21, line 187: some cpu does not support bf16,  could you check the behavior,  will it fallback to fp32 or just throw exception

neural_compressor/compression/pruner/pruners/pattern_lock.py
```diff
@@ -0,0 +1,59 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from .base import (register_pruner,
+                   PytorchBasePruner,
+                   KerasBasePruner)
+from ..patterns import get_pattern
+
+
+@register_pruner('pt_pattern_lock')
```
WeiweiZhang1, line 25: If pytorch pruners are prefixed with 'pt', all our existing example configs need to be updated.

neural_compressor/compression/pruner/pruners/__init__.py
```diff
@@ -0,0 +1,90 @@
...
+FRAMEWORK = {
+    'pytorch': 'pt',
+    'keras': 'keras'
+}
+
+
+def parse_valid_pruner_types():
+    """Get all valid pruner names."""
+    valid_pruner_types = []
+    for x in CRITERIA.keys():
```
WeiweiZhang1, line 39: There may be new pruner types added later that don't have a corresponding criterion, and this restriction may be a bit strict.

neural_compressor/compression/pruner/pruners/pattern_lock.py
n1ck-guo: No. This situation has been considered. For all these existing example, there is no different between this version and previous one. API change only happen under compression/pruner and for outside coding, the api is the same. The get_pruner function is exactly the same with previous version. Using parameter framework to return pytorch or keras pruners.

neural_compressor/compression/pruner/pruners/__init__.py
n1ck-guo: This func is used to show the valid pruner name. When we add a new criteria, we use register_criterion as a decorator and store the class name into a hyper parameter CRITERIA. Therefore, new criteria can be enter though CRITERIA.keys().

WeiweiZhang1 (approved):

n1ck-guo (approved):

PR #1051, opened 2023-06-30
Add quantized model download API for Neural Solution

Kaihui-intel:
## Type of Change

feature 
API changed or not: not

## Description

- add quantization model download api
- replace the quantized model path with download link of the response to the submission request
- Client downloads zip files(quantized_model.zip) without leaving temporary files on the server

## Dependency Change?

When installing `fastapi`, it will automatically install and update the `strarlette`.
`zipfile` is a built-in Python package.


yiliu30:
Please add usage to one of the E2E examples.

chensuyue (approved):

PR #1053, opened 2023-06-30
enable weight only quantization for language modeling

violetch24:
## Type of Change

enable weight only quantization for language modeling
model: gpt-j-6B
API not changed

local test


xin3he (approved):

xin3he:
Please trigger the extension test.

PR #1061, opened 2023-07-04
Fix ONNXRT diagnosis bug

yuwenzho:
## Type of Change

bug fix 
API not change

## Description

**Bug description:**
ONNXRT GPT2 static quantization with diagnosis=True failed with error "missing scale for tensor .."

**Optimized Solution:**
Get scale and zero point tensor recursivly.
Check name of dump tensor is not empty. 
Remove the node if it does not serve as the input or output of any other nodes.

tensors can be inspected correctly for ONNXRT GPT2 static quantization with diagnosis=True

ONNXRT GPT2 static quantization with diagnosis=True

## Dependency Change?

No


yuwenzho:
@bmyrcha Hi, I tested hf_albert-base-v2, hf_gpt2, hf_bart-large with diagnosis=True locally. Please help review. Thanks!

mengniwang95 (approved):

chensuyue (approved):

PR #1081, opened 2023-07-12
Fixed the mixed precision UT

yiliu30:
## Type of Change

bug fix
API changed or not: None

## Description

The execution duration was too short, resulting in a random optimized result. To address this, specify the execution time explicitly.

Pre-CI & stress-test

## Dependency Change?

None


chensuyue (approved):
Stress test pass.

PR #1083, opened 2023-07-13
support full range for 4 bit sym -8

xin3he:
## Type of Change

feature

## Description

support full range for 4 bit sym -8, use -scale to make sure fake sym quantization can be recovered.

UT pass

local test UT

## Dependency Change?

N/A


hshen14 (commented):
LGTM

PR #1085, opened 2023-07-13
fallback add when smoothquant

xin3he:
## Type of Change

enhancement

## Description

fallback add when smoothquant

add is fallbacked when smoothquant

local test UT and checked the stats table

## Dependency Change?

N/A

### DIFF
--- a/neural_compressor/adaptor/pytorch.py
+++ b/neural_compressor/adaptor/pytorch.py
@@ -2534,6 +2534,14 @@ class PyTorch_IPEXAdaptor(TemplateAdaptor):
 
         # For smoothquant optimized model
         recipe_cfgs = tune_cfg.get('recipe_cfgs', None)
+        ## fallback add when smoothquant is True
+        if recipe_cfgs and recipe_cfgs.get('smooth_quant', False):
+            for k, v in tune_cfg['op'].items():
+                if k[1] == 'add':
+                    tune_cfg['op'][k] = {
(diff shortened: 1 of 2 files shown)


hshen14 (approved):

changwangss:
I remember fallback add will cause large latency, is it right? @PenghuiCheng
if it is true, please remove "t5" out. @xin3he

hshen14:
> I remember fallback add will cause large latency for t5 and flan-t5, is it right? @PenghuiCheng if it is true, please remove "t5" out. @xin3he

@xin3he , by default, SmoothQuant will make Add FP32. Is it still possible to make Add INT8, if users specify in config through op-type? We should have such flexibility.

xin3he:
> ant will make Add FP32.

In you description, it's a fallback strategy which has been refused by @mengniwang95  and @yiliu30

xin3he:
Make user do the fallback add is still be best solution now.

PenghuiCheng:
t5, flan-t5 performance:
fallback "add", int8 model slower than fp32 about 4~5x
no fallback "add, int8 model faster than fp32 model about 1.2x.

In t5 and flan-t5 , there are many add ops.

hshen14:
Offline discussed with Yi, Xin, and Penghui. Yi will raise a new PR through the enhancement in strategy. Close it for now. Feel free to re-open if needed.

PR #1137, opened 2023-08-04
fix folding=None issue

xin3he:
## Type of Change

bug fix

## Description

folding=None is not applied auto setting, which caused opt-6.7b-sq is not applied with smoothquant.

model accuracy pass

local test


xin3he:
https://inteltf-jenk.sh.intel.com/view/lpot-validation/job/intel-lpot-validation-top-mr-extension/5239/

PR #1159, opened 2023-08-17
divide tf and pr pruning test in CI to detect dependency issue

chensuyue:
## Type of Change

validation 

## Description

divide tf and pr pruning test in CI to detect dependency issue

CI pass

CI
## Dependency Change?

No


chensuyue:
The failure of baseline is reasonable, because master doesn't have the new scripts.

PR #1207, opened 2023-09-01
Remove Version Limits for tf2oonx Examples

zehao-intel:
## Type of Change

bug fix

## Description

Remove version limits for tf2onnx examples because the dependency conflicts has been solved in the newest version.

Run tf2onnx examples with newest version of dependencies.

Extension test

## Dependency Change?

NA


chensuyue:
We should set   
`onnx>=1.13.0   
tensorflow>=2.11.1  `
to avoid CVEs.

zehao-intel:
> We should set `onnx>=1.13.0 tensorflow>=2.11.1 ` to avoid CVEs.

Solved.

PR #1216, opened 2023-09-05
Use TF as default Backend for ViT Example

zehao-intel:
## Type of Change

API not changed

## Description

Use tensorflow as the default backend for this example instead of ITEX.

PreCI, Extension.

## Dependency Change?

No


chensuyue:
Extension test pass

PR #1266, opened 2023-09-19
Reduce memory consumption in ONNXRT adaptor

yuwenzho:
## Type of Change

feature
API not change

## Description

Reduce memory consumption in ONNXRT adaptor:
- Delete unnecessry InferenceSession
- Refine check function for large models 

local test

## Dependency Change?

no

### DIFF
--- a/neural_compressor/model/onnx_model.py
+++ b/neural_compressor/model/onnx_model.py
@@ -18,6 +18,7 @@
 
 import logging
 import os
+import sys
 from pathlib import Path
 
 from neural_compressor.model.base_model import BaseModel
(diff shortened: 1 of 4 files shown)


neural_compressor/model/onnx_model.py
```diff
@@ -66,6 +61,23 @@ def __init__(self, model, **kwargs):
         self._get_graph_info()
         self._q_config = None
 
+    def check_large_model(self):
+        """Check model > 2GB."""
+        init_size = 0
+        for init in self._model.graph.initializer:
+            # if size of the initializer > 2GB, return True
+            try:
+                init_bytes = init.SerializeToString()
```
mengniwang95, line 70: if the initializer is saved to external file, can you still get its size with this method?

yuwenzho: Yes. `self._model = model if not isinstance(model, str) else onnx.load(model)` will load initializer data

mengniwang95: but if users pass loaded model object to this class, it will not execute onnx.load(model)

yuwenzho: update check of initializer data_location

hshen14 (approved):

mengniwang95 (approved):

PR #1269, opened 2023-09-20
Support GPTQ on ChatGLM2-6B

YIYANGCAI:
## Type of Change

bug fix

## Description

ChatGLM's source code, whose forwarding functions take several positional arguments, besides fixed "hidden_states". However, current GPTQ API only collects keyword arguments automatically. Thus, this pr is to add positional arguments automatic collection. 

Execute GPTQ quantization on ChatGLM2-6B automatically. 

PreCI

## Dependency Change?

no

### DIFF
--- a/examples/pytorch/nlp/huggingface_models/language-modeling/quantization/ptq_weight_only/run-gptq-llm.py
+++ b/examples/pytorch/nlp/huggingface_models/language-modeling/quantization/ptq_weight_only/run-gptq-llm.py
@@ -2,6 +2,7 @@ import sys
 sys.path.append("./")
 import math
 import time
+import re
 
 import torch
 import torch.nn as nn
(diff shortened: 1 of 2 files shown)


xin3he:
Is it possible that we use args and kwargs insteand of seperating hidden_state from them?
https://github.com/intel/neural-compressor/blob/7472254aeb425fda495b388cc52b4545099a18ee/neural_compressor/adaptor/torch_utils/awq.py#L457

wenhuach21 (approved):

xin3he (approved):

YIYANGCAI:
> Is it possible that we use args and kwargs insteand of seperating hidden_state from them? I think it would be more robust.
> 
> https://github.com/intel/neural-compressor/blob/7472254aeb425fda495b388cc52b4545099a18ee/neural_compressor/adaptor/torch_utils/awq.py#L457

Hi Xin, this is a good point, I will try it on multiple models.

PR #1334, opened 2023-10-19
Support wider range of models, especially CHN LLMs.

YIYANGCAI:
## Type of Change

Enhancement

## Description

- [x] Baichuan-7B

- [x] Baichuan2-7B

- [x] Qwen-7B

- [x] chatglm-6b (inference process should involve "with torch.autocast'cuda'")

- [x] chatglm2-6b

- [x] Mistral

- [x] phi-1

However, Baichuan, chatglm2-6b's evaluation currently not supported. \
Some unique characteristics for chatglm-6b and phi-1: 
- chatglm-6b **requires some specific token id (bos_token_id = 1300xx, etc.)** for successful inference. Thus, direct truncation on tokenized data is not suitable for this model. However, when untruncated sequences are input, the function runs properly.
- phi-1's torch.nn.Sequential object does not only contain the transformer stack but also embedding layers and final heads. Thus, I add a temporary function to trace the proper layers that can be quantized under GPTQ context.

### DIFF
--- a/neural_compressor/adaptor/torch_utils/gptq.py
+++ b/neural_compressor/adaptor/torch_utils/gptq.py
@@ -217,7 +217,6 @@ class GPTQuantizer(object):
         # model
         self.model = model
         # self.use_cache = self.model.config.use_cache
-        # import pdb;pdb.set_trace()
         self.gptq_related_blocks = trace_gptq_target_blocks(self.model)  # get the transformer block list above
         self.dtype = next(iter(self.model.parameters())).dtype
         log_quantizable_layers_per_transformer(self.gptq_related_blocks)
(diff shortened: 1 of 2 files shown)


xin3he (approved):

neural_compressor/adaptor/torch_utils/gptq.py
```diff
@@ -555,11 +581,17 @@ def tmp(_, inp, out):
             for layer_name in sub_layers:
                 handles.append(sub_layers[layer_name].register_forward_hook(add_batch(layer_name)))
             idx = self.cache_key_arguments.pop("i")
-            # import pdb;pdb.set_trace()
             for j in range(len(self.dataloader)):
                 cache_keyword_batch = self.gather_single_batch_from_dict(self.cache_key_arguments, j)
                 cache_positional_batch = self.gather_single_batch_from_list(self.cache_positional_arguments, j)
-                out = transformer_block(*cache_positional_batch, **cache_keyword_batch)[0]
+                if (
+                    hasattr(self.model.config, "_name_or_path") and "chatglm-6b" in self.model.config._name_or_path
```
wenhuach21, line 588: little ugly for production code

wenhuach21: how about change it to try catch?

hshen14 (approved):

PR #1372, opened 2023-11-03
Update new evaluation codes

YIYANGCAI:
Original evaluation codes in itrex failed in several CHN LLMs. After updating codes in itrex, such evaluation process in INC's example should also be improved.


hshen14:
Can we import itrex.eval to reuse the code?

YIYANGCAI:
> Can we import itrex.eval to reuse the code?

I have considered this once. Since we will call neural compressor in itrex, I am concerned that if we call itrex in a reverse manner, there might be some risks. Could @xin3he give some hints? Thanks!

xin3he (approved):

PR #1387, opened 2023-11-15
Support fp16 model to weight-only quantization for PyTorch framework

PenghuiCheng:
## Type of Change

feature
No API changed

## Description

Support fp16 model to weight-only quantization for PyTorch framework.

Quantization fp16 model successfully.

local tested


xin3he (approved):

xin3he (commented):
I think It's better to move the model.float() into the for loop`for name, m in model.named_modules():`

PenghuiCheng:
> I think It's better to move the model.float() into the for loop`for name, m in model.named_modules():`

yes, changed

chensuyue (approved):

PR #1431, opened 2023-11-30
change use_optimum_format=True and add bias

xin3he:
## Type of Change

bug fix

## Description

Optimum sets [bias=True](https://github.com/huggingface/optimum/blob/e2d3c8b6765e66c606b1cb59e3bb3e95453217e4/optimum/gptq/quantizer.py#L260) for QuantLinear when packing model. Here we follow this design of huggingface format for compatibility and set `use_hf_format=True` as default.
change argument name from hf to optimum.

UT pass


hshen14:
shall we specify the format, e.g., use_gptq_format? HF format sounds too general - how about AWQ and GGUF format? People also upload these formats to HF.

PenghuiCheng (approved):

hshen14 (approved):

PR #1451, opened 2023-12-08
fix pt notebook benchmark

violetch24:
## Type of Change
fix pt notebook benchmark


eduand-alvarez:
Hey there, these updates resolve some but not all issues. You can't load the model without a config file. Should we be using the optimum intel tooling to load this?

eduand-alvarez:
How do we load inc models that don't have config.json ?

violetch24:
> How do we load inc models that don't have config.json ?

If `save_results` path doesn't contain model configs, please use the following command instead:
```py
if args.input_model:
    from neural_compressor.utils.pytorch import load
    model = load(args.input_model, model, dataloader=eval_dataloader)
```
Modifications have been made in the latest commit.

eduand-alvarez:
Correct, with that change, the sample notebook runs as expected.

chensuyue (approved):

PR #1490, opened 2023-12-21
ONNX WOQ supports different dtypes

mengniwang95:
## Type of Change

enhancement

## Description

ONNX WOQ supports different dtypes

## Dependency Change?


hshen14 (approved):

neural_compressor/adaptor/ox_utils/weight_only.py
```diff
@@ -750,9 +764,9 @@ def awq_quantize(
             )
 
         session = (
-            ort.InferenceSession(model.model.SerializeToString(), so, providers=["CPUExecutionProvider"])
+            ort.InferenceSession(model.model.SerializeToString(), so, providers=ort.get_available_providers())
             if not model.is_large_model
-            else ort.InferenceSession(model.model_path + "_augment.onnx", so, providers=["CPUExecutionProvider"])
+            else ort.InferenceSession(model.model_path + "_augment.onnx", so, providers=ort.get_available_providers())
```
guotuofeng, line 769: In my testing with CPU version ORT, the get_available_providers will return ['AzureExecutionProvider', 'CPUExecutionProvider']. Not sure the extra Azure EP will cause some problem.

mengniwang95: got it, I updated the code with specified provider parameter

guotuofeng: do we need change https://github.com/intel/neural-compressor/blob/master/neural_compressor/model/model.py#L87? 

In my case, the default CPU EP would make the get_model_fwk_name return NA for ONNX model which has unsupported OPs in CPU

mengniwang95: could you provide more info like which op is not supported by CPU? As far as I know, ops don't supported by some EP will be fallback to CPU

guotuofeng: It is same with https://github.com/microsoft/onnxruntime/issues/16189, Trilu in CPU only support float/double inputs. CUDA Trilu support int_8/int_16/int_32.

guotuofeng: CPU trilu: https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/providers/cpu/tensor/trilu.cc#L106
cuda trilu: https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/core/providers/cuda/tensor/trilu_impl.cu#L38

yuwenzho (approved):

PR #1495, opened 2023-12-22
support models that need to be loaded using cuda

guotuofeng:
## Type of Change

Mistral model need to be loaded using CUDAExeuctionProvider. But the existing model loading only use CPUExecutionProvider.

## Description

This PR just enable the CUDA EP support.

## Dependency Change?

### DIFF
--- a/neural_compressor/adaptor/onnxrt.py
+++ b/neural_compressor/adaptor/onnxrt.py
@@ -980,11 +980,15 @@ class ONNXRUNTIMEAdaptor(Adaptor):
 
         if not model.is_large_model:
             sess = ort.InferenceSession(
-                model.model.SerializeToString(), sess_options, providers=["CPUExecutionProvider"]
+                model.model.SerializeToString(),
+                sess_options,
+                providers=["CUDAExecutionProvider", "CPUExecutionProvider"],
             )
         elif model.model_path is not None:  # pragma: no cover
             model.model = onnx.ModelProto()  # clean memory for large model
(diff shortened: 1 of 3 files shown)


yuwenzho:
Better to set `providers=onnxruntime.get_available_providers()`. CUDA EP may not be available on all devices.

guotuofeng:
> onnxruntime.get_available_providers()

It seems the onnxruntime.get_available_providers() doesn't work.
1. if the appropriated CUDA dependencies is not ready, the following errors will be raised.
image
2. On another hand, I install the CPU onnxruntime, and try to create InferenceSession by get_available_providers. then the following errors is raise:
image

guotuofeng:
> Better to set `providers=onnxruntime.get_available_providers()`. CUDA EP may not be available on all devices.

the default onnxruntime cpu version will have "AzureExecutionProvider" when invoking get_available_providers. Is that expected?

yuwenzho:
Hi @guotuofeng, By default, ONNX Runtime always places input(s) and output(s) on CPU. All model should be able to run on CPU EP. 
We plan to support CUDA EP for WOQ in this PR #1490

guotuofeng:
> Hi @guotuofeng, By default, ONNX Runtime always places input(s) and output(s) on CPU. All model should be able to run on CPU EP. We plan to support CUDA EP for WOQ in this PR #1490. We may not change code in `model.py` and `onnxrt.py`

So, should I close this PR since you guys will take care of it?

yuwenzho:
Support of CUDA EP for WOQ and related optimization of EP providers will be maintained in #1490

PR #1515, opened 2024-01-04
Use scm version

XuehaoSun:
## Type of Change

Use scm version when build from Source.
image

no

no

## Dependency Change?

no


setup.py
```diff
@@ -182,4 +181,6 @@ def fetch_requirements(path):
             "Topic :: Scientific/Engineering :: Artificial Intelligence",
             "License :: OSI Approved :: Apache Software License",
         ],
+        setup_requires=["setuptools_scm"],
+        use_scm_version={"version_scheme": "post-release"},
```
chensuyue, line 185: let's use dev release instead of post release.

setup.py
```diff
@@ -182,4 +193,9 @@ def fetch_requirements(path):
             "Topic :: Scientific/Engineering :: Artificial Intelligence",
             "License :: OSI Approved :: Apache Software License",
         ],
+        setup_requires=["setuptools_scm"],
```
chensuyue, line 196: `include_packages` doesn't work correctly with `setuptools_scm`, need further investigate.

chensuyue (approved):

PR #1539, opened 2024-01-12
WeightOnlyLinear keeps self.weight after recover

xin3he:
## Type of Change

user experience enhancement

## Description

WeightOnlyLinear keeps self.weight after recover due to recover is too slow.

UT pass

### DIFF
--- a/neural_compressor/adaptor/torch_utils/model_wrapper.py
+++ b/neural_compressor/adaptor/torch_utils/model_wrapper.py
@@ -259,7 +259,6 @@ class WeightOnlyLinear(torch.nn.Module):
                     dtype=self.float_type,
                 ).to(device),
             )
-            self.scales = self.scales.T
             self.register_buffer(
                 "qweight",
                 torch.zeros(
@@ -267,7 +266,6 @@ class WeightOnlyLinear(torch.nn.Module):
                     dtype=self.compression_dtype,
                 ).to(device),
             )
-            self.qweight = self.qweight.T
             self.register_buffer(
                 "qzeros",
                 torch.zeros(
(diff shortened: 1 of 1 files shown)


neural_compressor/adaptor/torch_utils/model_wrapper.py
```diff
@@ -329,6 +326,10 @@ def __init__(
             self.g_idx = None
 
     def pack(self, int_weight, scale, zp, bias, g_idx=None):
+        if self.use_optimum_format:
+            self.scales = self.scales.T
```
PenghuiCheng, line 330: Using in-place mode here, is it right?

xin3he: `Transpose` in Torch won't introduce a new tensor.

neural_compressor/adaptor/torch_utils/model_wrapper.py
```diff
@@ -468,12 +469,13 @@ def recover(self):
...
-            weight = weight.float()
-            self.bias = self.bias.float() if self.bias is not None else None
-        if level == DEBUG:
+        if not hasattr(self, "weight"):
+            weight = self.recover()
+            device = self.scales.device
+            if weight.dtype == torch.float16 and device.type == "cpu":
+                weight = weight.float()
+                self.bias = self.bias.float() if self.bias is not None else None
+        if True:  # keep reusing self.weight due to recover is too slow.
```
PenghuiCheng, line 478: if it is True, then "else" is not needed.

xin3he: keep it for debug usage.

changwangss (approved):

PenghuiCheng (approved):

PR #1558, opened 2024-01-22
Update joint architecture and quantization policy search with DyNAS-T example notebook

macsz:
There were two issues with the example notebook for joint architecture and quantization policy search with DyNAS-T.
1. Missing csv file with sample results.
2. CSV filename mismatch between cell that runs the search and the cell that plots results.

Both issues are resolved with this change.


macsz:
pre-commit CI complains `examples/notebook/dynas/ResNet50_Quantiation_Search_Supernet_NAS.ipynb:235: fo ==> of, for, to, do, go`. Problem is that line 235 is an encoded PNG image and there is not much that can be done about a typo within the encoding block.

chensuyue:
> pre-commit CI complains `examples/notebook/dynas/ResNet50_Quantiation_Search_Supernet_NAS.ipynb:235: fo ==> of, for, to, do, go`. Problem is that line 235 is an encoded PNG image and there is not much that can be done about a typo within the encoding block.

Ok, I will handle it.

PR #1563, opened 2024-01-23
refine config for auto-detecting backend and device

xin3he:
## Type of Change

enhancement

## Description

refine config for auto-detecting backend and device
- [x] torch
- [x] tensorflow
- [x] onnxrt

UT pass

### DIFF
--- a/neural_compressor/torch/quantization/config.py
+++ b/neural_compressor/torch/quantization/config.py
@@ -32,23 +32,10 @@ FRAMEWORK_NAME = "torch"
 DTYPE_RANGE = Union[torch.dtype, List[torch.dtype]]
 
 
-class Backend(Enum):
-    DEFAULT = "stock_pytorch"
-    IPEX = "ipex"
-
-
 class OperatorConfig(NamedTuple):
     config: BaseConfig
     operators: List[Union[str, Callable]]
-    backend: List[Backend]
     valid_func_list: List[Callable] = []
-
-
-# mapping the torch module type and functional operation type to string representations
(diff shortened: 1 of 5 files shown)


neural_compressor/torch/quantization/config.py
```diff
@@ -32,25 +32,12 @@
...
-class Backend(Enum):
-    DEFAULT = "stock_pytorch"
-    IPEX = "ipex"
-
-
 class OperatorConfig(NamedTuple):
     config: BaseConfig
     operators: List[Union[str, Callable]]
-    backend: List[Backend]
     valid_func_list: List[Callable] = []
```
yiliu30, line 39: In 3.x, the `supported_configs` is something similar to the `capability` in 2.x that defines the supported configs. Since various backends like `stock_pytorch` and `ipex` may have distinct capabilities, the `backend` in `OperatorConfig` serves to differentiate these supported configurations. Therefore, I think we shouldn't make changes to this code.

Regarding the automatic detection of the `backend` and `device`, I suggest modifying the config instance passed by the user automatically. For example:
``` python
# user_code.py
user_config = RTNConfig(backend='auto') # if the user did not set `backend`, default it `auto` 
# neural_compressor/torch/quantization/quantize.py
if user_config.backend == `auto` and ipex_is_available():
    user_config.backend = 'ipex' # override the user_config automatically
```

xin3he: INC should auto-detect the env before selecting implementation for different backends. It's an improvement comparing 2x, no worry.

PR #1605, opened 2024-02-05
update fp8 implementation, design and implement save&load 

xin3he:
## Type of Change

feature

## Description

update implementation and example for habana fp8.
- [x] refine example for less memory & better performance
- [x] remove workaround for old habana conda env
- [x] implement save & load interface

fp8 inference with lower memory usage and better performance.

local tested


neural_compressor/common/utils/save_load.py
```diff
@@ -0,0 +1,61 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import json
+import os
+
+
+def save_qconfig(config_mapping, qconfig_file_path):
```
yiliu30, line 22:
```suggestion
def save_config_mapping(config_mapping, file_path):
```

neural_compressor/common/utils/save_load.py
```diff
@@ -0,0 +1,61 @@
...
+    per_op_qconfig = {}
+    for (op_name, op_type), op_config in config_mapping.items():
+        value = {op_config.name: op_config.to_dict()}
+        per_op_qconfig[str((op_name, op_type))] = value
+
+    with open(qconfig_file_path, "w") as f:
+        json.dump(per_op_qconfig, f, indent=4)
+
+
+def load_qconfig(qconfig_file_path, config_name_mapping):
```
yiliu30, line 39:
```suggestion
def load_config_mapping(file_path, config_name_mapping):
```

setup.py
```diff
@@ -199,6 +212,8 @@ def get_build_version():
         include_package_data=True,
         package_data=package_data,
         install_requires=install_requires,
+        ext_modules=ext_modules,  # for fp8
```
yiliu30, line 215: This looks to be an FP8-only extension. How about using an environment variable(like `USE_FP8 = 1`, default is 0) to decide whether or not to build it? This way won't impose additional overhead on users who don't require FP8 functionality.

yiliu30 (approved):

setup.py
xin3he: good point

PR #1622, opened 2024-02-22
Support auto device for GPTQ and RTN

yiliu30:
## Type of Change

Feature: Support auto device for GPTQ and RTN
API changed: Introcude `FORCE_DEVICE=CPU`

## Description

The `Auto_Accelerator` components will automatically choose the device (`cuda`, `xpu`, `cpu`, etc.) based on their priority. 
We assume that users will not have two types of accelerated devices (such as `cuda` and `hpu`) on a single node.
In scenarios where a user has one accelerated device (like `cuda`) along with a CPU, and they want to use the CPU for testing or due to the limitations of CUDA memory, they can use **`FORCE_DEVICE=CPU`** to achieve this.

Some test on node with gpu may failed.

Pre-CI

## Dependency Change?
None


neural_compressor/torch/algorithms/weight_only/rtn.py
```diff
@@ -73,7 +74,14 @@ def rtn_quantize(
     Returns:
         model: fake quantized torch module
     """
-    device = "cpu"
+    device_name = kwargs.pop("device", "auto")
+    runtime_accelerator = auto_detect_accelerator(device_name)
+    device = runtime_accelerator.name()
```
xin3he, line 79: How about moving this device checking to `environ.py` so that all algos can call the result directly?

xin3he (approved):

neural_compressor/torch/algorithms/weight_only/rtn.py
yiliu30: Thanks, moved it into environ.py in [fixed UTs](https://github.com/intel/neural-compressor/pull/1622/commits/0186e67aaeb20412c8b0e70c0db46e8131701f15).

YIYANGCAI (approved):

PR #1662, opened 2024-03-11
support habana FP8 per channel quantization

xin3he:
## Type of Change

feature

## Description

- [x] support per channel quantization for higher accuracy
- [x] add observer registry for easy extension
- [x] dump scale_inv from observer to align with Habana Quantization Toolkit
- [x] move observer device to cpu to avoid program hangs

UT pass

local test


xin3he:
Will add UTs later

PR #1673, opened 2024-03-14
Add save&load API to SmoothQuant ipex model

violetch24:
## Type of Change

add save&load API to 3.x SmoothQuant and StaticQuant

ut passed


test/3x/torch/quantization/test_smooth_quant.py
```diff
@@ -110,11 +111,21 @@ def run_fn(model):
         q_model = quantize(fp32_model, quant_config=quant_config, run_fn=run_fn, example_inputs=example_inputs)
         assert q_model is not None, "Quantization failed!"
         inc_out = q_model(example_inputs)
-        q_model.save("saved")
+        # set a big atol to avoid random issue
+        assert torch.allclose(inc_out, ipex_out, atol=2e-02), "Unexpected result. Please double check."
+
+        from neural_compressor.torch.algorithms.smooth_quant import load, recover_model_from_json, save
+
+        save(q_model, "saved_results")
```
xin3he, line 119: Could we replace the save with ours?

test/3x/torch/quantization/test_smooth_quant.py
```diff
@@ -110,11 +111,21 @@ def run_fn(model):
...
-        loaded_model = torch.jit.load("saved")
+        loaded_model = load("saved_results")
         loaded_out = loaded_model(example_inputs)
-        assert torch.allclose(inc_out, ipex_out, atol=1e-05), "Unexpected result. Please double check."
+        # set a big atol to avoid random issue
+        assert torch.allclose(inc_out, loaded_out, atol=2e-02), "Unexpected result. Please double check."
 
-        assert torch.allclose(inc_out, loaded_out, atol=1e-05), "Unexpected result. Please double check."
+        fp32_model = copy.deepcopy(model)
+        ipex_model = recover_model_from_json(fp32_model, "ipex.json", example_inputs=example_inputs)
```
xin3he, line 128: How about adding the check of "saved_results/qconfig.json"?

xin3he (changes requested):

PR #1682, opened 2024-03-21
Fix export compressed model OOM

Kaihui-intel:
## Type of Change

bug fix

## Description

Fix OOM when export compressed model.
root cause: a) quantized weight `type` is not  an in-place operation. b)WeightOnlyLinear objects also occupy memory.
solution: a) return origin type

## Dependency Change?

### DIFF
--- a/neural_compressor/adaptor/torch_utils/weight_only.py
+++ b/neural_compressor/adaptor/torch_utils/weight_only.py
@@ -128,7 +128,7 @@ def qdq_weight_asym(weight, num_bits=4, quantile=1.0, return_int=False):
     weight.add_(zp)
     weight.clamp_(0, maxq)
     if return_int:
-        return weight.type(torch.uint8), scale.type(torch.float), zp.type(torch.uint8)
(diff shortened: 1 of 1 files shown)


neural_compressor/adaptor/torch_utils/weight_only.py
```diff
@@ -128,7 +128,7 @@ def qdq_weight_asym(weight, num_bits=4, quantile=1.0, return_int=False):
     weight.add_(zp)
     weight.clamp_(0, maxq)
     if return_int:
-        return weight.type(torch.uint8), scale.type(torch.float), zp.type(torch.uint8)
+        return weight, scale.type(torch.float), zp.type(torch.uint8)
```
xin3he, line 131: Could we remove all type call here?

xin3he (approved):

neural_compressor/adaptor/torch_utils/weight_only.py
Kaihui-intel: sure, removed, thanks.

PR #1705, opened 2024-04-01
Fix  RTN supported layer checking condition

Kaihui-intel:
## Type of Change

bug fix

## Description
Issue: `FalconLinear` is not quantized.
root cause: falcon model 's linear is built-in by itself, not used torch.nn.linear, the issue depends on INC return model
solution: change supported layers name string to isinstance checking.

## Dependency Change?


Kaihui-intel:
3x will fix on another PR

XuehaoSun (approved):

changwangss (approved):

PR #1710, opened 2024-04-02
WOQ models export workaround for autoround different device

changwangss:
## Type of Change

API changed or not

## Description

detail description

## Dependency Change?

### DIFF
--- a/neural_compressor/model/torch_model.py
+++ b/neural_compressor/model/torch_model.py
@@ -499,7 +499,7 @@ class PyTorchModel(PyTorchBaseModel):
 
         autoround_config = self.autoround_config if hasattr(self, "autoround_config") else {}
 
-        if gptq_config or (autoround_config and device=="xpu"):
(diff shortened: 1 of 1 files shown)


neural_compressor/model/torch_model.py
```diff
@@ -558,7 +558,7 @@ def export_compressed_model(
                 )
                 new_module.pack(int_weight, gptq_scale, gptq_zp, m.bias, gptq_perm)
                 set_module(self.model, k, new_module)
-        elif autoround_config:
+        elif autoround_config and (device == "cpu" or device == "auto"):
```
thuang6, line 561: does ITREX support inference with NV GPU?  if support,  does it go through device == "auto" path?

changwangss: we support NV GPU if cuda available, when device is "auto", it will detect cuda avaliable first and then use cpu device.

chensuyue:
Extension test pass

PR #1716, opened 2024-04-10
Update model ckpt download url in prepare_model.py

addf400:
## Type of Change

bug fix

API Changed

## Description

This pull request addresses an urgent security issue where a live Azure Storage SAS token was inadvertently exposed in the codebase. The exposed secret has been removed to prevent any unauthorized access to Microsoft's resources.

The file prepare_model.py within the onnxrt/image_recognition/beit/quantization/ptq_static directory previously contained a hardcoded Azure Storage SAS token, which is against best practices and could potentially compromise the security of our services.

The token has been revoked to ensure the security of our services. As an additional measure, I have replaced the URL with a placeholder to prompt the user to insert their own credentials.

The expected behavior after this PR is merged is that the code will no longer contain any live secrets, and users will be prompted to use their own credentials for Azure Storage access. The potential risk of not merging this PR is continued exposure of the live secret, which could lead to unauthorized access and manipulation of Microsoft's Azure resources.

The new BEiT ckpt download is accessible
## Dependency Change?

No new library dependencies were introduced or removed with this PR.

### DIFF
--- a/examples/onnxrt/image_recognition/beit/quantization/ptq_static/prepare_model.py
+++ b/examples/onnxrt/image_recognition/beit/quantization/ptq_static/prepare_model.py
@@ -6,7 +6,7 @@ from urllib import request
 from timm.models import create_model
 import beit_modeling_finetune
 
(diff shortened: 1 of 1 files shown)


donglixp:
The original model link is going to retire. We update a new link for the BEiT checkpoint.

chensuyue (approved):

PR #1718, opened 2024-04-10
fix tune_cfg issue for 3.x static quant

violetch24:
## Type of Change

bug fix

## Description

fixed tune_cfg loading problem for 3.x static quant API

UT passed

### DIFF
--- a/neural_compressor/torch/algorithms/static_quant/utility.py
+++ b/neural_compressor/torch/algorithms/static_quant/utility.py
@@ -159,7 +159,16 @@ def check_cfg_and_qconfig(user_cfg, cfgs, op_infos_from_cfgs, output_tensor_ids_
                         inc_scheme = inc_op_cfg["activation"]["scheme"]
                         inc_algorithm = inc_op_cfg["activation"]["algorithm"]
                         ipex_op_cfg["input_tensor_infos"] = input_tensor_infos
-                        activation_observer = generate_activation_observer(inc_scheme, inc_algorithm)
+                        if (
+                            "op_type" in ipex_op_cfg
+                            and ipex_op_cfg["op_type"] == "<class 'torch.nn.modules.linear.Linear'>"
+                        ):
+                            smooth_quant_enable = True
+                        else:
+                            smooth_quant_enable = False
+                        activation_observer = generate_activation_observer(
+                            inc_scheme, inc_algorithm, smooth_quant=False, smooth_quant_enable=smooth_quant_enable
+                        )
                         if inc_scheme == "sym":
                             input_tensor_infos[index]["force_dtype"] = "torch.qint8"
                         if inc_scheme == "asym":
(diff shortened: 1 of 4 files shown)


neural_compressor/torch/algorithms/static_quant/utility.py
```diff
@@ -63,57 +66,238 @@
...
+    """This is a helper method to generate an activation observer.
+
+    Args:
+        scheme (str): Quantization scheme to be used.
+        algorithm (str): What algorithm for computing the quantization parameters based on.
+
+    Returns:
+        An observer.
+    """
+    kl_activation_observer = {
```
yiliu30, line 220: Do these observer configurations already exist in IPEX? If so, kindly import them straight.

neural_compressor/torch/algorithms/static_quant/utility.py
```diff
@@ -63,57 +66,238 @@
...
+        op_infos = copy.deepcopy(op_infos_from_cfgs)
+        cfgs = check_cfg_and_qconfig(tune_cfg["op"], cfgs, op_infos, output_tensor_id_op_name, smooth_quant=False)
+
+        with open(ipex_config_path, "w") as write_f:
+            json.dump(cfgs, write_f, indent=4)
+        return None
+
+
+def check_cfg_and_qconfig(
+    tune_cfg, cfgs, op_infos_from_cfgs, output_tensor_ids_op_name, smooth_quant=False
```
yiliu30, line 134: The name `tune_cfg` isn't suitable because quantization not aware the tuning process. How about use `user_cfg`?

neural_compressor/torch/algorithms/static_quant/utility.py
```diff
@@ -63,57 +66,238 @@
...
+):  # pragma: no cover
+    """Check configs and quantization configs.
```
yiliu30, line 136: I'm a bit confused about the function name and its docstring. Could it be made clearer?

yiliu30 (approved):

Kaihui-intel:
https://github.com/intel/neural-compressor/blob/a29addc9c908b7dd15f71f04f3c1a1a08cdf7a04/neural_compressor/torch/algorithms/static_quant/static_quant.py#L97-L99
Why do we need to save the configuration twice in different ways? 
Perhaps the first one should be INC's config?

Kaihui-intel (approved):

violetch24:
> https://github.com/intel/neural-compressor/blob/a29addc9c908b7dd15f71f04f3c1a1a08cdf7a04/neural_compressor/torch/algorithms/static_quant/static_quant.py#L97-L99
> 
> 
> Why do we need to save the configuration twice in different ways?
> Perhaps the first one should be INC's config?



> https://github.com/intel/neural-compressor/blob/a29addc9c908b7dd15f71f04f3c1a1a08cdf7a04/neural_compressor/torch/algorithms/static_quant/static_quant.py#L97-L99
> 
> 
> Why do we need to save the configuration twice in different ways?
> Perhaps the first one should be INC's config?

`model.tune_cfg = json.load(f)` is to add a new attribute `tune_cfg` that includes configurations, and will be used in save API.

changwangss:
Could you remove the code about IPEX version lower than 1.12?

violetch24:
> Could you remove the code about IPEX version lower than 1.12?

Removed.

PR #1729, opened 2024-04-11
Fix case where pad_max_length set to None

echarlaix:
Set default value for the `pad_max_length` attribute to avoid the case where it's set to `None` as done here https://github.com/intel/neural-compressor/blob/ba479850d3365fa8fae145b1376f104211af8b66/neural_compressor/adaptor/pytorch.py#L4531 by default as it'll results in issue when comparing its value during quantization like done here https://github.com/intel/neural-compressor/blob/v2.5/neural_compressor/adaptor/torch_utils/gptq.py#L307

An other solution could be to modify `framework_specific_info["recipes"]`

### DIFF
--- a/neural_compressor/adaptor/torch_utils/gptq.py
+++ b/neural_compressor/adaptor/torch_utils/gptq.py
@@ -248,7 +248,7 @@ class GPTQuantizer(object):
 
         # dataloader
         self.use_max_length = use_max_length
-        self.pad_max_length = pad_max_length
+        self.pad_max_length = pad_max_length or 2048
         self.dataloader_original = dataloader
         self.dataloader = []
         self.nsamples = nsamples


xin3he:
Hi @echarlaix , I think this change is not so reasonable as GPTQ has default value 2048 for pad_max_length. If pad_max_length is None, it is passed by the user through the recipe configuration. So I think we should teach user set it to a number or not set it (using the default value). https://github.com/intel/neural-compressor/blob/master/docs/source/quantization_weight_only.md#examples

echarlaix:
Hi @xin3he, currently if a user doesn't specify `max_input_length` when creating an instance of `GPTQConfig` then quantization will fail as the default value for `max_input_length` is [None](https://github.com/intel/intel-extension-for-transformers/blob/4122f13de70dc6a208d4177eec73568f12b2f80e/intel_extension_for_transformers/transformers/utils/config.py#L706C9-L706C25) for `GPTQConfig` , which results in `pad_max_length` to be set to None  [here](https://github.com/intel/intel-extension-for-transformers/blob/4122f13de70dc6a208d4177eec73568f12b2f80e/intel_extension_for_transformers/transformers/llm/quantization/utils.py#L486) which is not something expected by [GPTQuantizer](https://github.com/intel/neural-compressor/blob/v2.5/neural_compressor/adaptor/torch_utils/gptq.py#L307). An other solution would be to update the default value for `GPTQConfig`

xin3he:
Thank you, @echarlaix , I got it.
In [transformers/quantization_config](https://github.com/huggingface/transformers/blob/2c66600c3fbbdcfc1610c5862c33bd19cf42859e/src/transformers/utils/quantization_config.py#L493), the default value of max_input_length is None, So here our gptq algorithm should be compatible with it since it is popular usage. Thanks again for your contribution.

PR #1740, opened 2024-04-19
map ipex op_name w/ pt op_name

violetch24:
## Type of Change

feature: add ipex op_name mapping
bug fix: op_type set local config

## Description
ipex op_name <-> eager model op_name mapping, for users to better use fallback by op_name for ipex
eg: 
image

ut


yiliu30 (approved):
LGTM, thanks!

test/3x/torch/quantization/test_static_quant.py
```diff
@@ -49,6 +49,21 @@ def test_static_quant_default(self):
         q_model = quantize(fp32_model, quant_config=quant_config, run_fn=run_fn, example_inputs=example_inputs)
         assert q_model is not None, "Quantization failed!"
 
+    @pytest.mark.skipif(not is_ipex_available(), reason="Requires IPEX")
+    def test_static_quant_fallback(self):
+        fp32_model = copy.deepcopy(self.fp32_model)
+        quant_config = get_default_static_config()
+        example_inputs = self.input
+        # fallback by op_type
+        quant_config.set_local(torch.nn.modules.linear.Linear, StaticQuantConfig(w_dtype="fp32", act_dtype="fp32"))
```
xin3he, line 58: Please use string 'Linear' instead of objective. Objective is more complex and cannot be saved directly as json.

neural_compressor/torch/quantization/config.py
```diff
@@ -923,7 +923,7 @@ def register_supported_configs(cls) -> List[OperatorConfig]:
     def get_model_info(model: torch.nn.Module, example_inputs) -> List[Tuple[str, Callable]]:
         from neural_compressor.torch.algorithms.smooth_quant import get_quantizable_ops_recursively
 
-        model_info, _, _, _ = get_quantizable_ops_recursively(model, example_inputs=example_inputs)
+        model_info, _, _, _, _ = get_quantizable_ops_recursively(model, example_inputs=example_inputs)
```
xin3he, line 926: I'm a little confused about line 821 and 926, which use different position to return model_info.

test/3x/torch/quantization/test_static_quant.py
```diff
@@ -49,6 +49,21 @@ def test_static_quant_default(self):
...
+        quant_config = get_default_static_config()
+        example_inputs = self.input
+        # fallback by op_type
+        quant_config.set_local(torch.nn.modules.linear.Linear, StaticQuantConfig(w_dtype="fp32", act_dtype="fp32"))
+        q_model = quantize(fp32_model, quant_config=quant_config, run_fn=run_fn, example_inputs=example_inputs)
+        assert q_model is not None, "Quantization failed!"
+
+        # fallback by op_name
+        quant_config.set_local("fc1", StaticQuantConfig(w_dtype="fp32", act_dtype="fp32"))
+        q_model = quantize(fp32_model, quant_config=quant_config, run_fn=run_fn, example_inputs=example_inputs)
```
xin3he, line 64: Do you have a check on LLMs, such as llama2?

xin3he (approved):

neural_compressor/torch/quantization/config.py
violetch24: This op_name mapping feature has not been added to SmoothQuant yet (might be added later), thus using `get_quantizable_ops_recursively` to return two different formats of model info: converted pt format for static quant, and the original ipex format for sq for now.

test/3x/torch/quantization/test_static_quant.py
xin3he: Aligned with @yiliu30, we will accept original torch class and method as input of `set_local` to avoid confusion with op_name and use its string name internally to map with IPEX op_type.

test/3x/torch/quantization/test_static_quant.py
violetch24: checked on gptj

PR #1742, opened 2024-04-19
Fix `weight_only`  algorithms import 

Kaihui-intel:
## Type of Change

bug fix

## Description

RTN is not related with auto_round, so we should make this dependency independent.

## Dependency Change?


yiliu30 (commented):
The `neural_compressor/torch/algorithms/weight_only/autoround.py` is for auto-round, it can import `auto_round` from anywhere in the code.

Importing these components lazily would be a better way?

https://github.com/intel/neural-compressor/blob/ba7564833ccf660688350235115ad9429efb078d/neural_compressor/torch/algorithms/weight_only/__init__.py#L15-L21



```python
autoround = LazyImport("neural_compressor.torch.algorithms.weight_only.autoround.autoround_quantize")
```

Kaihui-intel:
@yiliu30 , What do you think of using `LazyImport` in autoround.py?

Kaihui-intel:
@yiliu30 , What do you think of using Lazyimport in autoround.py?

> The `neural_compressor/torch/algorithms/weight_only/autoround.py` is for auto-round, it can import `auto_round` from anywhere in the code.
> 
> Importing these components lazily would be a better way?
> 
> https://github.com/intel/neural-compressor/blob/ba7564833ccf660688350235115ad9429efb078d/neural_compressor/torch/algorithms/weight_only/__init__.py#L15-L21
> 
> ```python
> autoround = LazyImport("neural_compressor.torch.algorithms.weight_only.autoround.autoround_quantize")
> ```

If `LazyImport` is utilized, the autonomy of the algorithm code cannot be ensured, as it triggers the inclusion of additional INC API code.

xin3he (approved):
LGTM

yiliu30 (approved):

PR #1743, opened 2024-04-19
auto detect available device when exporting

xin3he:
## Type of Change

feature

## Description

auto detect available packing device when exporting, priority: ["xpu", "cuda", "cpu"]

When XPU format is required and XPU device is not available, INC will select packing device from ["xpu", "cuda", "cpu"] automatically.

test on cuda device when using xpu argument.


changwangss (approved):

neural_compressor/model/torch_model.py
```diff
@@ -498,6 +498,18 @@ def export_compressed_model(
             gptq_config = self.gptq_config if hasattr(self, "gptq_config") else {}
 
         autoround_config = self.autoround_config if hasattr(self, "autoround_config") else {}
+        # check available device, priority: ["xpu", "cuda", "cpu"]
+        availiable_device = []
```
yiliu30, line 502: Are these device priorities in line with `auto_accelerator`? https://github.com/intel/neural-compressor/blob/master/neural_compressor/torch/utils/auto_accelerator.py
If yes, how about registering the `xpu` to as a new accelerator and get the device by https://github.com/intel/neural-compressor/blob/fdb5097907acae0e60834423c6c00323e73b962e/neural_compressor/torch/utils/auto_accelerator.py#L264

xin3he: This is a design for 2x. To avoid code injection, I think we should keep a separate `auto_accelerator` check.
I will add another PR to register `XPU` as a new accelerator for 3x. 
2x will keep this simple design as there is no such complex requirement.

xin3he: https://github.com/intel/neural-compressor/pull/1754

yiliu30 (approved):

xin3he:
Keep this branch exist since it is used by IPEX.

PR #1789, opened 2024-05-12
Fix type hints

yiliu30:
## Type of Change

API changed or not

## Description

detail description

## Dependency Change?


neural_compressor_ort/quantization/config.py
```diff
@@ -552,7 +554,7 @@ def get_model_info(self, model, *args, **kwargs):
 
 
 def get_all_config_set_from_config_registry() -> Union[BaseConfig, List[BaseConfig]]:
```
mengniwang95, line 556: only List[Type[BaseConfig]]?

yiliu30: Thanks, fixed!

PR #1793, opened 2024-05-15
Support Autotune FP16 Mix-precision on torch 3.0 new API

zehao-intel:
## Type of Change

feature

## Description

support fp16 by using torch.amp

PreCI

## Dependency Change?

No


neural_compressor/torch/algorithms/mix_precision/module_wrappers.py
```diff
@@ -0,0 +1,40 @@
...
+    """FP16Module Wrapper Class."""
+
+    def __init__(self, module, device):
+        """Init a FP16ModuleWrapper object."""
+        super(FP16ModuleWrapper, self).__init__()
+        self.add_module("module", module)
+        self.train(module.training)
+        self.device = device
+        # WA for TransformerEncoder to access its Linear's weights and bias
+        if isinstance(module, torch.nn.Linear):
```
yiliu30, line 32: How to handle other modules?

neural_compressor/torch/algorithms/mix_precision/module_wrappers.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+class FP16ModuleWrapper(torch.nn.Module):
+    """FP16Module Wrapper Class."""
+
+    def __init__(self, module, device):
+        """Init a FP16ModuleWrapper object."""
+        super(FP16ModuleWrapper, self).__init__()
+        self.add_module("module", module)
+        self.train(module.training)
```
yiliu30, line 29: Currently, we do not consider the training scenarios, kindly remove it.

neural_compressor/torch/algorithms/mix_precision/fp16_convert.py
```diff
@@ -0,0 +1,70 @@
...
+# limitations under the License.
+"""FP16 Convert for Torch Modules."""
+
+from typing import Dict, Tuple
+
+import torch
+
+from neural_compressor.common import logger
+from neural_compressor.torch.algorithms.mix_precision.module_wrappers import FP16ModuleWrapper
+from neural_compressor.torch.quantization import MixPrecisionConfig
```
yiliu30, line 25: Please do not import `xxx` from `neural_compressor.torch.quantization` into `neural_compressor.torch.algorithms`, it may cause the circular import.

test/3x/torch/test_autotune.py
```diff
@@ -332,6 +338,51 @@ def eval_acc_fn(model):
...
+
+    @reset_tuning_target
+    def test_autotune_mix_precision_set_op_name(self):
+        from neural_compressor.common.base_config import ComposableConfig, config_registry
+        from neural_compressor.torch.algorithms.mix_precision import HalfPrecisionModuleWrapper
+
+        def eval_acc_fn(model) -> float:
+            return 1.0
+
+        config = {
```
yiliu30, line 364: I think current `tuning_cinfig` only generate one config, please enhance it to generate a set of configs (> 1). A more reliable method to ensure the expected behavior is to verify whether the weight's dtype is `fp16` or not.

yiliu30 (approved):

test/3x/torch/test_autotune.py
zehao-intel: Hi @yiliu30 ，
In fact, I only expect to generate one config here. The local config is just for setting one Module to `fp32` instead of `fp16`.
And for the data type validation, I think the weight's `dtype` will still be `fp32` because the `torch.autocast` will not permanently convert this attribute to half-precision. The casting only happens when running `forward`(inside the `torch.autocast` context managers).

yiliu30: > In fact, I only expect to generate one config here. The local config is just for setting one Module to fp32 instead of fp16.

It's fine to write a UT that generates only one config. However, how about adding more UTs that generate more than one configs to verify the tuning behavior?

> And for the data type validation, ...

I understand, thanks.

zehao-intel: > It's fine to write a UT that generates only one config. However, how about adding more UTs that generate more than one configs to verify the tuning behavior?
Thanks for your suggestion. Yes, you are right. Maybe we could update this feature in a separate PR later.

PR #1801, opened 2024-05-17
freeze the pt backend peft version to avoid ut issue

chensuyue:
## Type of Change

others  

## Description

freeze the pt backend peft version to avoid ut issue


chensuyue:
@violetch24  please remove the limitation after fix the sq ut issue

PR #1816, opened 2024-05-24
fix 3.x IPEX examples failed with evaluate

violetch24:
## Type of Change

3.x example bug fix

pass extension test


changwangss:
Can you provide an offline model verification? eg: before/after smoothquant accuracy

PR #1821, opened 2024-05-28
3.x SQ autotune supports calib_func w/ capture input

violetch24:
## Type of Change
bug fix for 3.x sq autotune

## Description
Layer-wise & block-wise enable
Add ut check auto-tune
Check llm examples

## Dependency Change?

### DIFF
--- a/neural_compressor/torch/algorithms/smooth_quant/utility.py
+++ b/neural_compressor/torch/algorithms/smooth_quant/utility.py
@@ -25,6 +25,7 @@ import tqdm
 from packaging.version import Version
 
 from neural_compressor.torch.algorithms.static_quant import (
+    CpuInfo,
     TransformerBasedModelBlockPatternDetector,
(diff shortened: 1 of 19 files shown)


neural_compressor/torch/algorithms/smooth_quant/utility.py
```diff
@@ -412,6 +418,44 @@ def model_forward(model, dataloader, iters, device):  # pragma: no cover
...
+            super().__init__()
+            self.args_list = []
+            self.kwargs_list = []
+            self.orig_model = model
+            self.iters = 0
+            self.calib_num = calib_num
+
+        def forward(self, *args, **kwargs):
+            if self.iters >= self.calib_num:
+                raise ValueError
```
xin3he, line 447: We don't need raise ValueError now

xin3he (approved):

PR #1832, opened 2024-06-03
set automatic return_dict=False for GraphTrace

xin3he:
## Type of Change

bug fix

## Description

Currenly, AWQ UTs doesn't set return_dict/torchscript for transformers model, so the GraphTrace is always failed.
MulLinear will be inserted to help AWQ work, and is not necessary. 
Ber this PR, return_dict is automatically set and make AWQ ease-of-use. 

UT pass

### DIFF
--- a/neural_compressor/torch/algorithms/weight_only/utility.py
+++ b/neural_compressor/torch/algorithms/weight_only/utility.py
@@ -729,12 +729,11 @@ class GraphTrace:
         if orig_device != "cpu" and orig_device != "meta":  # pragma: no cover
             model = model.to("cpu")
             dummy_input = move_input_to_device(dummy_input, "cpu")
-        try:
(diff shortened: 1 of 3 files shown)


neural_compressor/torch/algorithms/weight_only/utility.py
```diff
@@ -729,6 +729,12 @@ def trace(self, model, dummy_input):
         if orig_device != "cpu" and orig_device != "meta":  # pragma: no cover
             model = model.to("cpu")
             dummy_input = move_input_to_device(dummy_input, "cpu")
+        try:
+            # set return_dict=False to help transformers model jit.trace success
+            orig_return_dict = model.config.return_dict
```
wenhuach21, line 734: why using try? better using if

wenhuach21 (approved):

PR #1838, opened 2024-06-04
Update torch mx example script

mengniwang95:
## Type of Change

example update

## Description

detail description

## Dependency Change?


xin3he (approved):

chensuyue:
If we want to add into weekly test, then require `run_quant.sh` and `run_benchmark.sh`, and add config into the 3x json config.

PR #1844, opened 2024-06-05
add examples

xin3he:
## Type of Change

example

## Description

- [ ] rename mx to mx_quant
- [ ] add dlrm static_quant on ipex backend


examples/.config/model_params_pytorch_3x.json
```diff
@@ -0,0 +1,11 @@
+{
+  "pytorch": {
```
chensuyue, line 2: The 3x Pytorch examples should use the same config file, so the 3x llm examples' config should also added here. 
The model list is: llama2_7b_gptq_int4,llama2_7b_gptq_int4_dq_bnb,llama2_7b_gptq_int4_dq_ggml,gpt_j_woq_rtn_int4,gpt_j_woq_rtn_int4_dq_bnb,gpt_j_woq_rtn_int4_dq_ggml,gpt_j_woq_gptq_int4,gpt_j_woq_gptq_int4_dq_bnb,gpt_j_woq_gptq_int4_dq_ggml,opt_125m_woq_gptq_int4,opt_125m_woq_gptq_int4_dq_bnb,opt_125m_woq_gptq_int4_dq_ggml  
opt_125m_ipex,opt_125m_ipex_sq,llama2_7b_ipex,llama2_7b_ipex_sq,gpt_j_ipex,gpt_j_ipex_sq

xin3he: We will rename the folder from `llm` to `weight_only` and  `smooth_quant`, so it won't be included in this PR.  
@Kaihui-intel @violetch24 please follow Suyue's requirement.

xin3he:
rely on bug fix: https://github.com/intel/neural-compressor/pull/1861

xin3he:
extension test: 
https://inteltf-jenk.sh.intel.com/job/intel-lpot-validation-top-pr-extension-3x-API/21/

neural_compressor/torch/quantization/autotune.py
```diff
@@ -93,17 +93,19 @@ def autotune(
...
-            # !!! Make sure to use deepcopy only when inplace is set to `True`.
-            q_model = quantize(
-                deepcopy(model),
-                quant_config=best_quant_config,
-                run_fn=run_fn,
-                run_args=run_args,
-                inplace=True,
-                example_inputs=example_inputs,
-            )
+            if trial_index == 0:  # recover the best q_model from previous results.
```
yiliu30, line 96: Good point, However, we might need to refine the condition a bit further. It would be more robust to check whether the `best_trial_index` is equal to the `last_trial_index`. If so, we can skip the recovery process. `trial_index == 0` is one such case.

xin3he: yeah, we can target it at next release

chensuyue (approved):

PR #1864, opened 2024-06-13
fix 3x ipex static quant regression

violetch24:
## Type of Change

bug fix

## Description
fix 3x ipex static quant regression
cannot fallback with op type name ('linear')
dump wrong op stats (no 'Linear&relu' op type)


xin3he (approved):

changwangss (approved):

xin3he:
https://inteltf-jenk.sh.intel.com/job/inc-validation-top-weekly-large-model-SPR-only-3x-API/54/

PR #1880, opened 2024-06-20
Support xpu device for 3.x ipex static

violetch24:
## Type of Change

feature
API not changed

## Description

add support to xpu device for 3.x ipex static

local test


neural_compressor/torch/quantization/config.py
```diff
@@ -995,6 +997,7 @@ def __init__(
         self.act_granularity = act_granularity
         self.act_algo = act_algo
         self.excluded_precisions = excluded_precisions
+        self.device = device
```
yiliu30, line 1000: I have a general question: should we infer the `device` from the model, or should we ask users to set it explicitly? 
If users have already placed the model on XPU using sth like `model.to(xpu)`, I believe we should infer it from the model by checking the device information of the model's parameters.

neural_compressor/torch/quantization/config.py
```diff
@@ -968,6 +968,7 @@ class StaticQuantConfig(BaseConfig):
         "act_granularity",
         "act_algo",
         "excluded_precisions",
+        "device",
```
yiliu30, line 971: Please only add the tunable parameter to the `params_list`.

neural_compressor/torch/quantization/config.py
```diff
@@ -1005,17 +1008,17 @@ def register_supported_configs(cls) -> List[OperatorConfig]:
         supported_configs.append(OperatorConfig(config=linear_static_config, operators=operators))
         cls.supported_configs = supported_configs
 
-    @staticmethod
-    def get_model_info_for_ipex(model: torch.nn.Module, example_inputs) -> List[Tuple[str, Callable]]:
+    def get_model_info_for_ipex(self, model: torch.nn.Module, example_inputs) -> List[Tuple[str, Callable]]:
```
yiliu30, line 1011: If we infer the `device` from the model, I think we can leave it as a static method.

yiliu30 (approved):

violetch24:
Will raise a PR with new design.

PR #1902, opened 2024-07-04
example update for 3.x ipex sq

violetch24:
## Type of Change

example refine

## Description

use TSModel for loading, similar to ITREX
add autotune in example for ipex sq


xin3he (approved):
LGTM

PR #1904, opened 2024-07-09
Get default config based on the auto-detect CPU type

yiliu30:
## Type of Change

feature
API changed or not: Yes

## Description
Need to auto-detect CPU and provide lightweight algorithm configuration as default config for client.
- [ ] Doc

Algo:
- [x] RTN
- [x] GPTQ
- [x] AUTO-ROUND

### Usage
```python
config_auto_detect = get_default_rtn_config()
config_for_client = get_default_rtn_config(processor_type="client")
config_for_server = get_default_rtn_config(processor_type="server")
```
### Test
```bash
pytest -sv ./test/3x/torch/test_config.py -k  TestQuantConfigBasedonProcessorType
```

## Dependency Change?


neural_compressor/torch/utils/utility.py
```diff
@@ -278,3 +280,61 @@ def get_model_device(model: torch.nn.Module):
...
+    Server = "Server"
+
+
+def detect_processor_type_based_on_hw():
+    """Detects the processor type based on the hardware configuration.
+
+    Returns:
+        ProcessorType: The detected processor type (Server or Client).
+    """
+    # Detect the processor type based on below conditions:
```
yiliu30, line 296: @xin3he @Kaihui-intel Here is the auto-detect logic. We can refine it if you have other suggestions.

xin3he: I'm not sure if this is correct. Also, maybe the number of cores is also a consideration.

xin3he (approved):

neural_compressor/torch/utils/utility.py
yiliu30: I enhanced the check a little bit and added more logs to inform the user of the detection result. If the detection result is wrong, the user can override it explicitly。

Kaihui-intel (approved):

PR #1911, opened 2024-07-11
implement TorchBaseConfig

xin3he:
## Type of Change

implement TorchBaseConfig

## Description

Re-write func `_get_op_name_op_type_config` in BaseConfig to fallback op_type with string.
Reason:
- [x] IPEX backend has some special op_types, such as `Linear&add`, which cannot be set to class.
- [x] keep the same behavior as INC 2.x for torch backend.

UTs pass

### DIFF
--- a/test/3x/torch/quantization/test_static_quant.py
+++ b/test/3x/torch/quantization/test_static_quant.py
@@ -76,7 +76,7 @@ class TestStaticQuant:
         quant_config = get_default_static_config()
         example_inputs = self.input
         # fallback by op_type
-        quant_config.set_local(torch.nn.Linear, StaticQuantConfig(w_dtype="fp32", act_dtype="fp32"))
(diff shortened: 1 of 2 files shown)


test/3x/torch/quantization/test_static_quant.py
```diff
@@ -76,7 +76,7 @@ def test_static_quant_fallback(self):
         quant_config = get_default_static_config()
         example_inputs = self.input
         # fallback by op_type
-        quant_config.set_local(torch.nn.Linear, StaticQuantConfig(w_dtype="fp32", act_dtype="fp32"))
+        quant_config.set_local([torch.nn.Linear, "Linear&add"], StaticQuantConfig(w_dtype="fp32", act_dtype="fp32"))
```
yiliu30, line 79: Please add some UTs to ensure that all `Linear&add` are fallbacked as expected.

xin3he: It's added already. The message is dumped in log and verified manually.

violetch24 (approved):

Kaihui-intel (approved):

PR #1915, opened 2024-07-12
[For Review Only] Release Notes for v3.0

thuang6:
v3.0 Release notes draft, [render version](https://github.com/intel/neural-compressor/blob/thuang6/release-notes-3.0/release-notes-v3.0.md)

- [x] FP8 quantization support after Habana HQT Integration commits upstream back to INC
- [x] INT4 model loading support after Habana commits upstream back to INC
- [x] Remain PRs for client side quantization improvements
- [x] Remain PRs for examples 
- [x] Remain PRs for documents 
- [ ] Multimodal model support in Highlights and examples if it is fitted into v3.0  (dropped)


release-notes-v3.0.md
```diff
@@ -0,0 +1,109 @@
...
+- Features
+- Improvements
+- Examples
+- Bug Fixes
+- Documentations
+- External Contributions
+- Validated Configurations
+
+**Highlights**
+ - FP8 quantization and INT4 model loading support on [Intel® Gaudi® AI accelerator](https://habana.ai/products/gaudi2/)
```
hshen14, line 13: How about FP8 quantization and weight-only INT4 quantization support on xxx?

release-notes-v3.0.md
```diff
@@ -0,0 +1,109 @@
...
+- Improvements
+- Examples
+- Bug Fixes
+- Documentations
+- External Contributions
+- Validated Configurations
+
+**Highlights**
+ - FP8 quantization and INT4 model loading support on [Intel® Gaudi® AI accelerator](https://habana.ai/products/gaudi2/) 
+ - Framework extension API (3.x API) for quantization, mixed-precision and benchmarking
```
hshen14, line 14: Do we need to mention 3.x here?

release-notes-v3.0.md
thuang6: That's my previous draft, but I revised to this.  not sure if we have a clear alignment to claim INT4 WOQ support on xxx? since we only covers RTN/AutoRound without formal release test

release-notes-v3.0.md
thuang6: I found "3.x" was used a lot in PR and also example file path, so I added this to indicate it is the same thing as framework extension API.  For release notes, I am ok to remove it if we believe it is kind of informal.

thuang6: removed

PR #1916, opened 2024-07-12
Support xpu for ipex static quant

violetch24:
## Type of Change

feature and example

## Description

Support xpu for ipex static quant
- [x] support quantized on xpu in current API
- [ ] save & load using config mapping
- [x] enable op_name and op_type fallback
- [x] add xpu related ut

ut pass

### DIFF
--- a/neural_compressor/torch/algorithms/static_quant/static_quant.py
+++ b/neural_compressor/torch/algorithms/static_quant/static_quant.py
@@ -33,11 +33,13 @@ from packaging.version import Version
 
 from neural_compressor.torch.algorithms import Quantizer
 from neural_compressor.torch.utils import logger
+from neural_compressor.torch.utils.auto_accelerator import auto_detect_accelerator
 
 from .utility import (
     CpuInfo,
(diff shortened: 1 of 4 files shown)


neural_compressor/torch/algorithms/static_quant/save_load.py
```diff
@@ -32,9 +32,17 @@ def save(model, output_dir="./saved_results"):
...
+        model.ori_save(qmodel_file_path)
+        with open(qconfig_file_path, "w") as f:
+            json.dump(model.tune_cfg, f, indent=4)
+    else:
+        from neural_compressor.common.utils import save_config_mapping
+
+        save_config_mapping(model.qconfig, qconfig_file_path)
+        # MethodType 'save' not in state_dict
+        del model.save
+        torch.save(model.state_dict(), qmodel_file_path)
```
xin3he, line 45: This save is not correct, jit model has it's own save method and we can reuse in both CPU and XPU situations.

xin3he (approved):

PR #1953, opened 2024-07-26
Fix to_json_file function in BaseConfig

yuwenzho:
## Type of Change

bug fix
API changed or not

## Description

Fix to_json_file function in BaseConfig

local test

## Dependency Change?

no


yiliu30 (approved):
Thanks!

chensuyue:
Will fixed in internal repo first.

PR #1954, opened 2024-07-26
Fix itrex qbits nf4/int8 training core dumped issue

Kaihui-intel:
## Type of Change

bug fix

## Description

TBB not installed when use `parallel=True`

`numba -s` has a section __Threading Layer Information__ that reports on the availability of threading layers in the current environment.
https://numba.readthedocs.io/en/stable/user/threading-layer.html#which-threading-layers-are-available

solution:
`pip install tbb`
Docker may need to set up an environment
`export LD_LIBRARY_PATH=/usr/local/lib/:$LD_LIBRARY_PATH`

## Dependency Change?


yiliu30 (approved):

yiliu30:
> export LD_LIBRARY_PATH=/usr/local/lib/

-> `export LD_LIBRARY_PATH=/usr/local/lib/:$LD_LIBRARY_PATH`

chensuyue:
We should add the description here: https://github.com/intel/neural-compressor/blob/master/docs/source/faq.md

chensuyue (approved):

PR #1985, opened 2024-08-23
add hasattr check for torch fp8 dtype

xin3he:
## Type of Change

bug fix

## Description

dtype initialization in global is not compatible with old torch version.
fix it with `hasattr` check

UT pass with old torch version


thuang6 (approved):

thuang6:
how about adding this fix to patch release to avoid failure in old Torch version @chensuyue

chensuyue (approved):

PR #1993, opened 2024-09-03
 Remove the save of gptq config

Kaihui-intel:
## Type of Change

feature 

## Description

Before 3.x, INC returned qdq model and then packed it with gptq-config. Now, INC directly return a packed model.

## Dependency Change?


yiliu30:
Please update the PR description to provide more background on why we need to remove it.

xin3he (approved):

yiliu30 (approved):

PR #2002, opened 2024-09-12
enable auto_round format export

WeiweiZhang1:
## Type of Change

feature
API not changed


Kaihui-intel (approved):
In order to pass CI, it may be necessary to switch to the version of Autoround and ask @XuehaoSun  for help

XuehaoSun (approved):

PR #2014, opened 2024-09-27
[Doc] Add autoround EMNLP24 paper to pub list

thuang6:
## Type of Change

documentation


wenhuach21:
it's on finding track not the main track, shall we add this?

wenhuach21 (approved):

thuang6:
> it's on finding track not the main track, shall we add this?

It is also part of EMNLP, IMO, it is ok to add it.

PR #2042, opened 2024-10-24
fix GPTQ oom issue on HPU

xin3he:
## Type of Change

bug fix

## Description

some memory is not released due to lack of `torch.hpu.synchronize()` and `gc.collect()`
Making sure the process is working as below:
image

llama2-70b works on one Gaudi2 card.

### DIFF
--- a/neural_compressor/torch/algorithms/weight_only/gptq.py
+++ b/neural_compressor/torch/algorithms/weight_only/gptq.py
@@ -607,8 +607,8 @@ class RAWGPTQuantizer(object):
                 for j in range(batch_num):
                     cache_keyword_batch = self.gather_single_batch_from_dict(self.cache_key_arguments, j)
                     cache_positional_batch = self.gather_single_batch_from_list(self.cache_positional_arguments, j)
-                    accelerator.mark_step()
                     out = transformer_block(*cache_positional_batch, **cache_keyword_batch)
+                    accelerator.synchronize()
                     out = self.track_hidden_states(out)
                 self.cache_key_arguments["batch_num"] = batch_num
                 for h in handles:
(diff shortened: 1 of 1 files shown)


neural_compressor/torch/algorithms/weight_only/gptq.py
```diff
@@ -690,13 +698,13 @@ def tmp(_, inp, out):
                 # Step 2.6: export to compressed model
                 for layer_name in sequential_layers:
                     weight_config_this_layer = self.get_layer_config(self.get_full_layer_name(layer_name, block_idx))
-                    gptq_scale = gptq_config[self.get_full_layer_name(layer_name, block_idx)]["scale"]
+                    gptq_scale = gptq_config[self.get_full_layer_name(layer_name, block_idx)]["scale"].cpu()
```
Kaihui-intel, line 701: Why do we give up supporting quantization on GPUs?

Kaihui-intel (approved):

neural_compressor/torch/algorithms/weight_only/gptq.py
xin3he: This is not correct logic.  weight is already moved to cpu three line above. Logic is like below:
image

xin3he: If we move the pack logic to device, HPU will allocate unexpected memory (not resolved). So I move it to cpu for all device since the CPU already provides objective speed.

n1ck-guo (approved):

xin3he:
@ulivne Please aware this. I will cherry-pick this to Habana-internal

PR #2044, opened 2024-10-24
add buckets setting for lm_eval

xin3he:
## Type of Change

feature

## Description

add buckets argument for lm_eval, it is used by HPU to limit static input length range.
By default, `buckets: Optional[list] = [64, 128, 256, 512, 1024, 2048, 4096]`

If default buckets are not enough to hold input length, it will cause error.
If used buckets is too many, it will take too much memory. (Each used bucket will cause the model to be copied)

### DIFF
--- a/neural_compressor/evaluation/lm_eval/models/huggingface.py
+++ b/neural_compressor/evaluation/lm_eval/models/huggingface.py
@@ -116,11 +116,13 @@ class HFLM(TemplateLM):
         peft: Optional[str] = None,
         autogptq: Optional[Union[bool, str]] = False,
         pad_to_buckets: Optional[Union[bool]] = False,
+        buckets: Optional[list] = [64, 128, 256, 512, 1024, 2048],
         model_format: Optional[str] = "torch",
         **kwargs,
     ) -> None:
         super().__init__()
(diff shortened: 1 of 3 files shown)


Kaihui-intel (approved):

changwangss (approved):

neural_compressor/evaluation/lm_eval/models/huggingface.py
```diff
@@ -116,11 +116,14 @@ def __init__(
         peft: Optional[str] = None,
         autogptq: Optional[Union[bool, str]] = False,
         pad_to_buckets: Optional[Union[bool]] = False,
+        buckets: Optional[list] = [32, 64, 128, 256, 512, 1024, 2048, 4096],
```
thuang6, line 119: why is 8192 removed?  result holding of too much memory?

thuang6 (approved):

neural_compressor/evaluation/lm_eval/models/huggingface.py
xin3he: yes, and it's not used. 4096 is used by wikitext

PR #2055, opened 2024-11-04
Update torch config kwargs

Kaihui-intel:
## Type of Change

bug fix

## Description

To avoid loading error if new config parameter is introduced in later release.

### DIFF
--- a/neural_compressor/torch/quantization/config.py
+++ b/neural_compressor/torch/quantization/config.py
@@ -147,6 +147,7 @@ class RTNConfig(TorchBaseConfig):
         quant_lm_head: bool = False,
         # Tuning space
         white_list: Optional[List[OP_NAME_OR_MODULE_TYPE]] = DEFAULT_WHITE_LIST,
+        **kwargs,
     ):
         """Init RTN weight-only quantization config.
 
@@ -380,6 +381,7 @@ class GPTQConfig(TorchBaseConfig):
         true_sequential: bool = False,
         # Tuning space
         white_list: Optional[List[OP_NAME_OR_MODULE_TYPE]] = DEFAULT_WHITE_LIST,
+        **kwargs,
     ):
         """Init RTN weight-only quantization config.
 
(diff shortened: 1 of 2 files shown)


neural_compressor/torch/quantization/config.py
```diff
@@ -380,6 +381,7 @@ def __init__(
         true_sequential: bool = False,
         # Tuning space
         white_list: Optional[List[OP_NAME_OR_MODULE_TYPE]] = DEFAULT_WHITE_LIST,
+        **kwargs,
     ):
         """Init RTN weight-only quantization config.
```
xin3he, line 386:
```suggestion
        """Init GPTQ weight-only quantization config.
```

xin3he (approved):
Nice, thanks.

neural_compressor/torch/quantization/config.py
```diff
@@ -147,6 +147,7 @@ def __init__(
         quant_lm_head: bool = False,
         # Tuning space
         white_list: Optional[List[OP_NAME_OR_MODULE_TYPE]] = DEFAULT_WHITE_LIST,
+        **kwargs,
```
yiliu30, line 150: Do we need to pass it to `TorchBaseConfig`?

yiliu30 (approved):

neural_compressor/torch/quantization/config.py
Kaihui-intel: Not need it for now. We are using the `__init__` method from `BaseConfig` . 
If necessary, we can add the `__init__` method to `TorchBaseConfig`, what's your opinion?

yiliu30: There's no need to add `__init__` for `TorchBaseConfig` until it has a specific logical difference from `BaseConfig`.

If there is common logic to handle `kwargs`, it would be better to pass it to `BaseConfig`.

Kaihui-intel: I agree with you.

PR #2109, opened 2025-01-09
[dont' merge] test removing code

xin3he:


xin3he:
By discussing with @Kaihui-intel, we can remove it now.
It was used for `torch.save(model，'model_path')`

PR #2113, opened 2025-01-24
Add intel-extension-for-pytorch to Transformers-like API requirements

XuehaoSun:
## Type of Change

API changed or not

## Description

detail description

## Dependency Change?

### DIFF
--- a/examples/3.x_api/pytorch/nlp/huggingface_models/language-modeling/quantization/transformers/weight_only/text-generation/requirements_cpu_woq.txt
+++ b/examples/3.x_api/pytorch/nlp/huggingface_models/language-modeling/quantization/transformers/weight_only/text-generation/requirements_cpu_woq.txt
@@ -15,3 +15,4 @@ lm-eval==0.4.3
 huggingface_hub
 numba
 tbb
+intel-extension-for-pytorch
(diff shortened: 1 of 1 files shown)


examples/3.x_api/pytorch/nlp/huggingface_models/language-modeling/quantization/transformers/weight_only/text-generation/requirements_cpu_woq.txt
```diff
@@ -15,3 +15,4 @@ lm-eval==0.4.3
 huggingface_hub
 numba
 tbb
+intel-extension-for-pytorch
```
thuang6, line 18: do we have version dependency or min version dependency?

XuehaoSun: > do we have version dependency or min version dependency?

Yes, it depends on the torch version, we need to update the torch version at the same time.
BTW is it necessary to freeze the torch version?

thuang6: I think it is ok to keep current verified torch version unless we have reason to move.  for IPEX even without specifying the version.  IPEX 2.4 version will be used to match torch, am I right?

XuehaoSun: > I think it is ok to keep current verified torch version unless we have reason to move. for IPEX even without specifying the version. IPEX 2.4 version will be used to match torch, am I right?

NO, it will install the latest IPEX version

thuang6: if that's the case, I would suggest keep IPEX and torch version matched.

thuang6 (approved):

Kaihui-intel (approved):
why

PR #2133, opened 2025-02-28
fix NotImplementedError: get_type is not implemented

xin3he:
## Type of Change

bug fix

## Description

blocking print after prepare.


nirda7 (changes requested):
This is not a valid fix.
Please take a look at my comment in here:
https://jira.habana-labs.com/browse/SW-220806

you cannot have abstract methods or even class methods that throw exceptions in the base class that none of the classes that inherits from it doesn't implement.

 If you want the refactor to stay - you should implement it in all classes in helper_modules.py
 and only then make sure all tests with their new patched modules also implement them...

yiliu30:
I think we should implement `get_type` and `get_module_type` for the existing patched modules in `helper_module.py` as well. Please refer the ticket for more details https://jira.habana-labs.com/browse/SW-220806

yiliu30 (approved):

nirda7 (approved):

PR #2143, opened 2025-03-10
Add `update_g_idx` flag for setting qweight&g_idx

Kaihui-intel:
## Type of Change

bug fix

## Description
```
error:
 [ERROR][modeling_auto.py:128] index 133 is out of bounds for dimension 0 with size 128
 [ERROR][modeling_auto.py:129] Saved low bit model loading failed, please check your model.
```
add `update_g_idx` flag to avoid secondary value updates.
note:
waiting for the release of ipex-gpu v2.7 ...
related PR: https://github.com/intel-innersource/frameworks.ai.pytorch.ipex-gpu/pull/5411

results:
meta-llama/Meta-Llama-3-8B-Instruct on PVC
|    Tasks     |Version|Filter|n-shot|  Metric  |   |Value |   |Stderr|
|--------------|------:|------|-----:|----------|---|-----:|---|-----:|
|lambada_openai|      1|none  |     0|acc       |↑  |0.7093|±  |0.0063|
|              |       |none  |     0|perplexity|↓  |3.5535|±  |0.0997|

## Dependency Change?

### DIFF
--- a/neural_compressor/transformers/quantization/utils.py
+++ b/neural_compressor/transformers/quantization/utils.py
@@ -238,7 +238,7 @@ def _replace_linear(
                             dtype=torch.int32,
                             device=torch.device(device),
                         )
-                    
+
                     # Note: update_g_idx is only applicable for ipex versions >=2.7
                     model._modules[name].set_weights_bias(
                         module.qweight.data if hasattr(module, "qweight") else weight,
@@ -247,7 +247,7 @@ def _replace_linear(
                             {"update_g_idx": not empty_weights}
                             if "update_g_idx" in model._modules[name].set_weights_bias.__code__.co_varnames
                             else {}
-                        )
+                        ),
                     )
                 else:
                     raise Exception("{} device Unsupported weight only quantization!".format(device))
--- a/neural_compressor/transformers/quantization/utils.py
+++ b/neural_compressor/transformers/quantization/utils.py
@@ -21,6 +21,7 @@ import re
 import types
 
 from datasets import load_dataset
+from packaging.version import Version
 
 from neural_compressor.common.utils import LazyImport, logger
 from neural_compressor.torch.algorithms.weight_only.modules import INCWeightOnlyLinear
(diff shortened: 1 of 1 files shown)


thuang6:
I saw IPEX PR has been merged. have we validated it? We do not need to wait for IPEX-XPU v2.7.  If there is issue, we need to raise before their release.

neural_compressor/transformers/quantization/utils.py
```diff
@@ -241,6 +241,7 @@ def _replace_linear(
                     model._modules[name].set_weights_bias(
                         module.qweight.data if hasattr(module, "qweight") else weight,
                         None if module.bias is None else module.bias.data,
+                        update_g_idx=not empty_weights,
```
thuang6, line 244: Better consider compatibility,  for old version IPEX without this flag, should not break. we can provide clear error msg to indicate this is not supported due to feature gap in this version for duplicated case.  if non-duplicated case, we should get it work without this flag.

Kaihui-intel: The issue has been resolved to ensure compatibility with older versions of IPEX. For versions without the required flag, the code now works without breaking. https://github.com/intel/neural-compressor/pull/2143/commits/8c0cc57b843f3a13b02c71be8cc8fb9ba18c72df

XuehaoSun (approved):

neural_compressor/transformers/quantization/utils.py
```diff
@@ -238,10 +241,17 @@ def _replace_linear(
                             dtype=torch.int32,
                             device=torch.device(device),
                         )
-                    model._modules[name].set_weights_bias(
-                        module.qweight.data if hasattr(module, "qweight") else weight,
-                        None if module.bias is None else module.bias.data,
-                    )
+                    if ipex_version >= Version("2.7"):
```
thuang6, line 244: is it possible to check interface itself support this arg instead of chekc IPEX version directly?  current approach looks not very robust

Kaihui-intel: Thanks, this can be implemented. I will submit another PR to replace the current code.

thuang6 (approved):

PR #2149, opened 2025-03-24
fix vulnerability

xin3he:


neural_compressor/utils/load_huggingface.py
```diff
@@ -96,33 +96,8 @@ def from_pretrained(cls, model_name_or_path: str, **kwargs) -> torch.nn.Module:
...
-                model_class._keys_to_ignore_on_load_unexpected = quantized_keys_to_ignore_on_load
-            else:
-                model_class._keys_to_ignore_on_load_unexpected.extend(quantized_keys_to_ignore_on_load)
-            missing_keys_to_ignore_on_load = [r"weight", r"bias"]
-            if keys_to_ignore_on_load_missing is None:
-                model_class._keys_to_ignore_on_load_missing = missing_keys_to_ignore_on_load
-            else:  # pragma: no cover
-                model_class._keys_to_ignore_on_load_missing.extend(missing_keys_to_ignore_on_load)
-
+            # only show logs of error level, since keys_to_ignore_on_load_unexpected is not working without specific model_class
```
thuang6, line 99: not understand the fix.   the issue is in line 86, why handle it here?

xin3he: Those codes are useless after changing the model_class.

thuang6 (approved):

PR #2157, opened 2025-03-31
[SW-219134]: Remove legacy import in init.

feng-intel:
To fix sw-219134. Remove legacy import in init.


thuang6:
issue was resolved in installation part.  this PR is not valid

PR #2171, opened 2025-04-14
suit transformers>=4.51

xin3he:


neural_compressor/torch/algorithms/weight_only/save_load.py
```diff
@@ -21,6 +21,8 @@
 import tempfile
 
 import torch
+import transformers
```
Kaihui-intel, line 24: `if is_transformers_imported():` or put it in function

neural_compressor/torch/algorithms/weight_only/save_load.py
```diff
@@ -846,7 +848,11 @@ def _init_hf_model(self, model_class, config):
 
             dtype_orig = model_class._set_default_torch_dtype(torch_dtype)
 
-        init_contexts = [no_init_weights(_enable=_fast_init)]
+        init_contexts = (
+            [no_init_weights(_enable=_fast_init)]
+            if parse(transformers.__version__) < parse("4.51")
```
Kaihui-intel, line 853: Suggest checking interface itself support this arg instead of check transformers version directly. Current approach looks not very robust.

Kaihui-intel (approved):

neural_compressor/torch/algorithms/weight_only/save_load.py
xin3he: discussed offline, better to keep this.

XuehaoSun (approved):

PR #2179, opened 2025-04-22
Add `add_bos_token` for Llama3 evaluation

Kaihui-intel:
## Type of Change

bug fix 

## Description

model: meta-llama/Llama-3.1-8B-Instruct
`add_bos_token` default is False.
If the model was trained or fine-tuned with a BOS token, this may lead to incorrect results.

## Dependency Change?


xin3he (approved):
good catch!

XuehaoSun:
image

XuehaoSun (approved):

PR #2203, opened 2025-05-08
Freeze setuptools==69.5.1

XuehaoSun:
## Type of Change

API changed or not

## Description

detail description

## Dependency Change?

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,3 +12,4 @@ pyyaml
 requests
 schema
 scikit-learn
+setuptools==69.5.1


xin3he (approved):

requirements.txt
```diff
@@ -12,3 +12,4 @@ pyyaml
 requests
 schema
 scikit-learn
+setuptools==69.5.1
```
thuang6, line 15: do we must freeze this version?  or can we use <= ?

thuang6:
I saw latest setuptools version is 80.3.1, while we try to freeze to 69.5.1.  Why did we find this issue recently instead of long ago?

XuehaoSun:
> I saw latest setuptools version is 80.3.1, while we try to freeze to 69.5.1. Why did we find this issue recently instead of long ago?

Yes, and the old version will have CVE, this package is updated very frequently, I have to close this PR and no longer freeze it

PR #2208, opened 2025-05-12
support llama2-70b gptq

xin3he:
## Type of Change

example enhancement
 
## Description

leverage mmap to support 70b gptq with limited host memory


xin3he:
Will commit to internal repo first

PR #2240, opened 2025-07-16
add CPU FP8 QDQ doc

mengniwang95:
## Type of Change

document

## Description

add CPU FP8 QDQ doc

depend on https://github.com/intel/neural-compressor/pull/2238


XuehaoSun (approved):

thuang6 (approved):
LGTM,  in future, FP8 mode likely needs to be supported for CPU after low precision op is ready

PR #2274, opened 2025-09-02
Adapt transformers 4.56.0

Kaihui-intel:
## Type of Change

adapt transformers 4.56.0

## Description

transformers==4.56.0 changed the import path, we now need to use transformers.pytorch_utils.Conv1D instead of transformers.modeling_utils.Conv1D
1. Use `transformers.pytorch_utils.Conv1D` instead of `transformers.modeling_utils.Conv1D`
2. use `_set_default_dtype` instead of `_set_default_torch_dtype`
3. fix pt2e ut: from transformers.integrations.executorch import export_with_dynamic_cache

## Dependency Change?

### DIFF
--- a/neural_compressor/transformers/quantization/utils.py
+++ b/neural_compressor/transformers/quantization/utils.py
@@ -579,7 +579,7 @@ def convert_to_quantized_model(model, config, device="cpu", for_inference=True):
             set_nontext_module_config(model, to_quant_block_names, config)
 
             for n, m in model.named_modules():
-                if isinstance(m, torch.nn.Linear) or isinstance(m, transformers.modeling_utils.Conv1D):
+                if isinstance(m, torch.nn.Linear) or isinstance(m, transformers.pytorch_utils.Conv1D):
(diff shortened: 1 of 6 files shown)


neural_compressor/transformers/quantization/utils.py
```diff
@@ -579,7 +579,7 @@ def set_nontext_module_config(model, to_quant_block_names, config):
             set_nontext_module_config(model, to_quant_block_names, config)
 
             for n, m in model.named_modules():
-                if isinstance(m, torch.nn.Linear) or isinstance(m, transformers.modeling_utils.Conv1D):
+                if isinstance(m, torch.nn.Linear) or isinstance(m, transformers.pytorch_utils.Conv1D):
```
yiliu30, line 582: Does that work for previous transformers versions?
If not, please import `Conv1D` based on the transformers version.

Kaihui-intel: It work. We should import `_set_default_dtype` based on the transformers version. Thanks for the reminder.

yiliu30 (approved):

PR #2291, opened 2025-09-23
Adapt AutoRound v0.8.0 [For MLLM]

Kaihui-intel:
## Type of Change

Remove AutoRoundMLLM() API

## Description

detail description

## Dependency Change?


Kaihui-intel:
Waiting for https://github.com/intel/auto-round/pull/857 to be merged into v0.8.0rc branch.

neural_compressor/torch/quantization/config.py
yiliu30: nit: please update the default value in docstring as well.

neural_compressor/torch/quantization/config.py
```diff
@@ -1096,6 +1105,18 @@ def register_supported_configs(cls) -> List[OperatorConfig]:
...
         cls.supported_configs = supported_configs
 
+    def get_params_dict(self):
+        """Get a dictionary containing the parameters and their values for the current instance.
+
+        Returns:
+            A dictionary containing the parameters and their values.
+        """
+        result = dict()
+        for param, value in self.__dict__.items():
```
yiliu30, line 1115: Please add a note to clarify why we are skipping `tokenizer` here.

neural_compressor/transformers/quantization/utils.py
```diff
@@ -528,11 +530,14 @@ def convert_to_quantized_model(model, config, device="cpu", for_inference=True):
                 bs=config.batch_size,
                 nsamples=config.n_samples,
             )
+
+        device_map = get_accelerator().current_device_name()
```
yiliu30, line 534: The result of `get_accelerator().current_device_name()` is `cuda:x`, does this align with how the `device_map` is designed in the auto-round?

test/3x/torch/quantization/weight_only/test_autoround.py
```diff
@@ -295,27 +300,77 @@ def test_mllm(self):
     @pytest.mark.skipif(not ct_installed, reason="The compressed-tensors module is not installed.")
     @pytest.mark.parametrize("scheme", ["MXFP4", "NVFP4"])
     def test_scheme(self, scheme):
-        fp32_model = copy.deepcopy(self.gptj)
+        # INC API
```
yiliu30, line 301: It would be better to extend the test to all `PRESET_SCHEMES`.
https://github.com/intel/auto-round/blob/main/auto_round/schemes.py

neural_compressor/transformers/quantization/utils.py
Kaihui-intel: The `device_map` here is only for the Transformers-like API and is only applicable to `xpu` and `cpu`. It will be determined based on the `INC_TARGET_DEVICE` set.

PR #2361, opened 2025-12-11
add OoTPatchedModuleFusedSDPA

yangulei:
## Motivation
Current implementation in vLLM will call FusedSDPA with a huge attention mask, which leads to bad performance. This PR introduced three implementations to get better performance.

## Usage
Three environment variables are introduced to control the implementation:
- `VLLM_HPU_FSDPA_SLICE_SEQ_LEN_THLD`: `int`, the threshold for `kv_len` (=q_len+prefix_len) to apply the implementations, defaults to `8192`, set to a value less equal to 0 to disable slicing.
- `VLLM_HPU_FSDPA_SLICE_CHUNK_SIZE`: `int`, chunk size for the slicing in the implementation, defaults to `4096`.
- `VLLM_HPU_FSDPA_SLICE_IMPL`: `str` with choices in `['split_kv', 'slice_causal', 'slice_qkv']`, used to select the implementations, defaults to `slice_qkv`.

## Implementations
For a FusedSDPA with `q_len=11525` and `prefix_len=10752`. The lengths will be padded and be truncated respectively to `q_len=16384` and `prefix_len=8192` before calling the FusedSDPA. The full attention mask is shown bellow, in which the Gray parts steads for the values to be masked out.
attention_mask

### Notation
The following images include rectangles with three colors:
- `rgb(255,0,0)`: `is_causal=False` and `attn_mask is not None`
- `rgb(255,255,0)`: `is_causal=True` and `attn_mask=None`
- `rgb(255,0,255)`: `is_causal=False` and `attn_mask=None`

### The original implementation
The original implementation pass the full attention mask and set `is_causal=False` and `valid_seq_len=None`, which results in bad TPC/MME pipeline. The implementation could be visualized as the following image.
slicing_schedule_fp8_origin

### The `SplitKV` implementation
This implementation call the FusedSPDA twice for the prefix part and causal part respectively. And do not pass `attn_mask` for the prefix part thus gives better performance.
slicing_schedule_fp8_SplitKV

### The `SliceCausal` implementation
This implementation further slice the causal part into smaller chunks as illustrated in the following image.
slicing_schedule_fp8_SliceCausal

### The `SliceQKV` implementation
This implementation further slice the prefix part as shown below.
slicing_schedule_fp8_SliceQKV

### DIFF
--- a/neural_compressor/torch/algorithms/fp8_quant/_quant_common/patched_helper_modules.py
+++ b/neural_compressor/torch/algorithms/fp8_quant/_quant_common/patched_helper_modules.py
@@ -109,16 +109,16 @@ class OoTPatchedVllmMixtureOfExpertsOpFP8(INCPatchedVllmMixtureOfExpertsOpFP8):
 class OoTPatchedModuleFusedSDPA(INCPatchedModuleFusedSDPA):
     def __init__(self, mod, parent, mod_extra_config, *args, **kwargs):
         super().__init__(mod, parent, mod_extra_config, *args, **kwargs)
-        self.qkv_slice_thld = int(os.getenv("PT_HPU_QKV_SLICE_SEQ_LEN_THLD", 4096))
+        self.qkv_slice_thld = int(os.getenv("VLLM_HPU_FSDPA_SLICE_SEQ_LEN_THLD", 4096))
         if self.qkv_slice_thld > 0:
-            self.qkv_chunk_size = int(os.getenv("VLLM_FUSEDSDPA_QKV_SLICE_CHUNK_SIZE", self.qkv_slice_thld))
+            self.qkv_chunk_size = int(os.getenv("VLLM_HPU_FSDPA_SLICE_CHUNK_SIZE", self.qkv_slice_thld))
 
         impl_mapping = {
             'split_kv': self.fp8_fsdpa_split_kv,
             'slice_causal': self.fp8_fsdpa_slice_causal,
             'slice_qkv': self.fp8_fsdpa_slice_qkv,
         }
-        qkv_slice_impl = os.getenv("PT_HPU_QKV_SLICE_IMPL", 'slice_qkv').lower()
+        qkv_slice_impl = os.getenv("VLLM_HPU_FSDPA_SLICE_IMPL", 'slice_qkv').lower()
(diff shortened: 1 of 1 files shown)


yangulei:
@yiliu30 @Wei-Lin-Intel @czhu15
Please help to review, thanks!

neural_compressor/torch/algorithms/fp8_quant/_quant_common/patched_helper_modules.py
```diff
@@ -89,9 +92,374 @@ def forward_quant(
...
+        prefix_out = self.dequant_output(prefix_out).to(torch.float32)
+
+        # calculate the causal part
+        causal_k = k[..., prefix_len:, :]
+        causal_v = v[..., prefix_len:, :]
+        causal_mask = attn_mask[..., -q_len:]
+        causal_res = self.fp8_fsdpa_fwd(q, causal_k, causal_v, causal_mask, dropout_p, scale, False, softmax_mode)
+        causal_out, causal_m, causal_linv = (gqa_output_reshape(x) if gqa else x for x in causal_res[:3])
+        causal_m = causal_m.to(torch.float32)
+        causal_linv = causal_linv.to(torch.float32) * (128.0 if softmax_mode != "fp32" else 1.0)
```
Wei-Lin-Intel: Only `fast` mode requires `* 128.0`, for `fp32` and `None` modes, scale is 1.0

yangulei: Done, thx!

yangulei:
> LGTM Please add usage in PR desc :)

Done.

czhu15:
So this solution only truncate the prefix, not padding?

yangulei:
> So this solution only truncate the prefix, not padding?

This PR is used by the `deepseek_r1` branch with chunked-prefill enabled. No padding for the prefix context.

yangulei:
> LGTM. Pls check with INC team on the license description though.

It's added by pre-commit.

czhu15:
> > LGTM. Pls check with INC team on the license description though.
> 
> It's added by pre-commit.
Leave to @yiliu30 for confirmation.

PR #2378, opened 2026-01-05
Enable QKV slicing for causal FusedSDPA in OoTPatchedModuleFusedSDPA

yangulei:
Set `VLLM_HPU_FSDPA_SLICE_CAUSAL="true"` to enable. Refer to https://github.com/intel/neural-compressor/pull/2361 for more details.


yangulei:
> LGTM One minor comment. Not sure if we can combine fp8_apc_fsdpa_slice_qkv and fp8_causal_fsdpa_slice_qkv in one function with extra argument like is_apc. Since there is quite a lot common code in these 2 functions.

Good catch! Actually, I tried a version that merged the causal and APC cases, but the logic for setting the value of `is_causal`, for using mask or not, and for slicing the `attn_mask` or create a causal one becomes complicated and confusing. So I changed back to separated function calls.

PR #2448, opened 2026-04-15
Fix activation scale inf issue for const_weight and const_scale

qgao007:
## Type of Change

bug fix

## Description

A lot of activation scale of inf when using one sample for calibration. For such layer, we should skip when it's marked _not_quantized, but it actually store value, due to constant weight initialization. 

resolved constant weight giberrish output issue. No foreseeable risk. 

It has been tested using example_old.py file. 

## Dependency Change?

No dependency change.

### DIFF
--- a/neural_compressor/jax/quantization/layers_static.py
+++ b/neural_compressor/jax/quantization/layers_static.py
@@ -243,6 +243,18 @@ class StaticQDQLayer(SaveableLayerMixin, keras.layers.Layer):
             None: Cleans up observers and sets quantized call.
         """
         self._tracker.unlock()
+        if not self._is_quantized:
+            # Clean up observer only if it exists
+            if hasattr(self, "input_observer"):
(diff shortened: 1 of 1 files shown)


neural_compressor/jax/quantization/layers_static.py
```diff
@@ -243,6 +243,18 @@ def post_quantization_cleanup(self):
...
+            if hasattr(self, "input_observer"):
+                if hasattr(self, "_layers") and self.input_observer in self._layers:
+                    self._layers.remove(self.input_observer)
+            # Set call to pass-through/original
+            if hasattr(self, "call"):
+                # pass through
+                pass
+            self._const_variables = []
+            self._tracker.lock()
+            return
```
anko-intel, line 257: Early return just disable converting weight and scale to const (attributes)  in line 264 and later

anko-intel (changes requested):
Added code just disable const_weight and const_scale feature

PR #2459, opened 2026-04-30
 [JAX] Add Gemma3 quantization test

wpietka:
## Type of Change
Test

## Description
Adds a new tests which quantize gemma3 model and run it with both text-only and image containing prompt

E2E validation of Gemma3 output

pytest  test/jax/test_gemma3_model.py

## Dependency Change?
No dependencies changed


test/jax/test_gemma3_model.py
```diff
@@ -0,0 +1,110 @@
...
+        keras.saving.save_model(gemma_q, save_path)
+        gemma_q = keras.saving.load_model(save_path)
+
+    answer = gemma_q.generate("Answer what is the capital city of England. ", max_length=20, strip_prompt=True)
+    print("Gemma answer: ", {answer})
+    assert "The capital city of England is London" in answer
+
+
+@pytest.mark.parametrize("dynamic", [True, False], ids=["dynamic=True", "dynamic=False"])
+def test_image_recognition(colva_beach_sq, model_dtype, quantization_dtype, dynamic):
```
anko-intel: gemma3_instruct_4b-v1 is originally given in bfloat, so using float doesn't increase accuracy and should reduce significantly reduce time consumed by test, please use bfloat here

wpietka: Done

test/jax/test_gemma3.py
```diff
@@ -0,0 +1,107 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
```
anko-intel, line 2: missing copyright

test/jax/test_gemma3_model.py
```diff
@@ -0,0 +1,107 @@
...
+def test_text_prompt(random_string, quantization_dtype, dynamic):
+    model_dtype = "float32"
+    gemma = load_model_from_preset(Gemma3CausalLM, "gemma3_instruct_270m", model_dtype)
+
+    def calib_fn(model):
+        _ = model.generate(random_string, max_length=100)
+
+    if dynamic:
+        config = DynamicQuantConfig(weight_dtype=quantization_dtype, activation_dtype=quantization_dtype)
+        gemma_q = quantize_model(gemma, config, None)
```
anko-intel: we can omit third parameter

test/jax/test_gemma3.py
```diff
@@ -0,0 +1,107 @@
...
+    else:
+        config = StaticQuantConfig(
+            weight_dtype=quantization_dtype, activation_dtype=quantization_dtype, const_scale=True, const_weight=True
+        )
+        gemma_q = quantize_model(gemma, config, calib_fn)
+
+    with tempfile.TemporaryDirectory() as tmpdir:
+        save_path = os.path.join(tmpdir, "gemma3_quantized.keras")
+        keras.saving.save_model(gemma_q, save_path)
+        gemma_q = keras.saving.load_model(save_path)
```
anko-intel: As we save the model it will be good to load it and use the loaded version

test/jax/test_gemma3_model.py
```diff
@@ -0,0 +1,107 @@
...
+    else:
+        config = StaticQuantConfig(
+            weight_dtype=quantization_dtype, activation_dtype=quantization_dtype, const_scale=True, const_weight=True
+        )
+        gemma_q = quantize_model(gemma, config, calib_fn)
+
+    with tempfile.TemporaryDirectory() as tmpdir:
+        save_path = os.path.join(tmpdir, "gemma3_quantized.keras")
+        keras.saving.save_model(gemma_q, save_path)
```
anko-intel: to test different path than in test above it will be good to use save_to_preset

test/jax/test_gemma3_model.py
```diff
@@ -0,0 +1,107 @@
...
+        gemma_q = keras.saving.load_model(save_path)
+
+    answer = gemma_q.generate("Answer what is the capital city of England. ", max_length=20, strip_prompt=True)
+    print("Gemma answer: ", {answer})
+    assert "The capital city of England is London" in answer
```
anko-intel: I think it is too strict check

wpietka: It works 100% of a time, so I think we should stick to this version. If at some point this starts failing that would indicate that either original model was modified or accuracy was lost during quantization

test/jax/test_gemma3.py
wpietka: This is exactly what the last line does

test/jax/test_gemma3.py
wpietka: Added

test/jax/test_gemma3_model.py
wpietka: Removed

test/jax/test_gemma3_model.py
wpietka: Changed to use preset

test/jax/test_gemma3.py
```diff
@@ -0,0 +1,125 @@
+#!/usr/bin/env python
```
anko-intel, line 1: maybe rename  file to test_gemma3.py

test/jax/test_gemma3_model.py
wpietka: Actually changed that to "London" in answer

test/jax/test_gemma3.py
wpietka: Done

test/jax/test_gemma3.py
```diff
@@ -0,0 +1,125 @@
...
+        gemma_q = quantize_model(gemma, config)
+    else:
+        config = StaticQuantConfig(
+            weight_dtype=quantization_dtype, activation_dtype=quantization_dtype, const_scale=True, const_weight=True
+        )
+        gemma_q = quantize_model(gemma, config, calib_fn)
+
+    with tempfile.TemporaryDirectory() as tmpdir:
+        save_path = Path(os.path.join(tmpdir, "gemma3_quantized"))
+        if save_path.exists():
```
anko-intel: as Temporary Directory() is used, it should never happened? What was wrong to insert it here ?

test/jax/test_gemma3.py
```diff
@@ -0,0 +1,125 @@
...
+            shutil.rmtree(save_path)
+        save_path.mkdir(parents=False)
```
anko-intel: this ( I think ) is done internally in save_to_preset()

test/jax/test_gemma3.py
```diff
@@ -0,0 +1,125 @@
...
+        gemma_q.save_to_preset(save_path)
+        gemma_q = Gemma3CausalLM.from_preset(str(save_path), dtype=model_dtype)
```
anko-intel: I know that it is formally ok, as previous values of gemma_q will be deleted, but I will prefer do it explicitly with new name for loaded model toe emphasize that it is loaded quantize model

anko-intel (changes requested):
I think we have to find a way to make this test less time consuming without loosing coverage/common sense

test/jax/test_gemma3.py
wpietka: Done

test/jax/test_gemma3.py
wpietka: You're right, removed

test/jax/test_gemma3.py
wpietka: It's dead code essentially, removed

wpietka:
> I think we have to find a way to make this test less time consuming without loosing coverage/common sense

I changed consts to false for image_recognition, and decreased max tokens from 500 to 400. Test now takes 10 minutes less

anko-intel (approved):
LGTM
add [JAX] prefix in the title before merge

test/jax/test_gemma3.py
```diff
@@ -0,0 +1,127 @@
...
+    else:
+        config = StaticQuantConfig(
+            weight_dtype=quantization_dtype, activation_dtype=quantization_dtype, const_scale=False, const_weight=False
+        )
+        gemma_q = quantize_model(gemma, config, calib_fn)
+
+    with tempfile.TemporaryDirectory() as tmpdir:
+        save_path = Path(os.path.join(tmpdir, "gemma3_quantized"))
+        gemma_q.save_to_preset(save_path)
+        gemma_q_loaded = Gemma3CausalLM.from_preset(str(save_path), dtype=model_dtype)
```
bkowalskiINTEL: Do we need to wrap the path in the Path object? We extract the string anyway, so I think it is redundant

wpietka: True, it's not needed anymore, removed

PR #2460, opened 2026-04-30
[JAX] Properly handle incomplete calibration for static quantization

wpietka:
## Type of Change
Bug fix

## Description
Currently when calibration function for static quantization is provided with incomplete input sample - not activating all model layers during calibration - certain layers are effectively broken, due to scales being set to 0.

This change detects if quantization failed for given layer and recovers original call() methods and weights. Hence the layer will behave exactly as if it was not quantized at all

Output provided by quantized model will be correctly even if quantization fails for certain layers

pytest  test/jax/test_gemma3_model.py::test_static_quantization_with_incomplete_calibration

## Dependency Change?
No dependencies changed


test/jax/test_gemma3_model.py
```diff
@@ -108,3 +108,38 @@ def calib_fn(model):
...
+    random_string, colva_beach_sq, model_dtype, quantization_dtype
+):
+    gemma = load_model_from_preset(Gemma3CausalLM, "gemma3_instruct_4b-v1", model_dtype)
+
+    # Run calibration without image in input, so vision layer won't activate during calibration
+    def calib_text_fn(model):
+        _ = model.generate(random_string, max_length=100)
+
+    config = StaticQuantConfig(
+        weight_dtype=quantization_dtype, activation_dtype=quantization_dtype, const_scale=True, const_weight=True
```
anko-intel: as across the tests in this file we always have const_scale=True, it will be good to test the opposite  setting here or in the other place, You can do it in the test which consumes more time ( as I assume in such short test compilation takes longer for const weights than reduce of time for inference due to constant usage - please check)
```suggestion
        weight_dtype=quantization_dtype, activation_dtype=quantization_dtype, const_scale=False, const_weight=False
```

neural_compressor/jax/quantization/layers_static.py
```diff
@@ -234,6 +234,21 @@ def convert(self):
...
+
+        Returns:
+            None: Prepares layer for loading scales and weights
+        """
+        self._tracker.unlock()
+        if self._is_quantized:
+            self.call = self.call_asymmetric if self._is_asymmetric else self.call_symmetric
+            self._trainable_variables[:] = [v for v in self._trainable_variables if v.name != "kernel"]
+        else:
+            self.call = self.call_no_quant
```
anko-intel: The first part is already done in post_quantization_cleanup
```suggestion
        if not self._is_quantized:
            self.call = self.call_no_quant
```

anko-intel (changes requested):
I would like to avoid adding so many load_own_variables_preprocess
Lets discuss it offline, Probably we can postpone post_quantization_cleanup actions into load_own_variables

test/jax/test_gemma3_model.py
wpietka: Done, made test 5 minutes quicker

neural_compressor/jax/quantization/layers_dynamic.py
```diff
@@ -83,6 +83,7 @@ def __init__(self, name, activation_dtype, dtype="float32", asymmetric=False):
         super().__init__(name=name, dtype=dtype)
         self.activation_dtype = activation_dtype
         self._is_asymmetric = asymmetric
+        self._is_quantized = True
```
anko-intel: it is only required to have during load_own_variables? Not used here at all ?

anko-intel (approved):
It looks very well, thanks

neural_compressor/jax/quantization/layers_dynamic.py
wpietka: Yes, I've added this variable for all dynamic layers so I can load _is_quantized from store without checking if it's there

anko-intel (approved):

neural_compressor/jax/quantization/layers_static.py
anko-intel, line 715: missing _is_quantized setting for q_qdq

anko-intel (changes requested):

neural_compressor/jax/quantization/layers_static.py
wpietka: Fixed

anko-intel (approved):
LGTM

bkowalskiINTEL (approved):
LGTM

PR #2462, opened 2026-05-06
Dev/bkowalsk/jax conv2d quantization

bkowalskiINTEL:
This PR adds Conv2D quantization support for Keras/JAX.
It inherits most of the logic from Dense/EinsumDense, but a minor tweak was needed in the *call* method, because for Conv2D it doesn't accept the *training* argument.


anko-intel (approved):
LGTM

PR #2466, opened 2026-05-12
[JAX] Use const scale in MHA after softmax

anko-intel:
## Type of Change

feature

## Description

 For tensor quantized inside MultiHeadAttention after softmax const range of values can be assumed. This way calibration or min.max finding for dynamic quantization is not required for this tensor.

## Dependency Change?

### DIFF
--- a/neural_compressor/jax/quantization/layers_dynamic.py
+++ b/neural_compressor/jax/quantization/layers_dynamic.py
@@ -68,7 +68,7 @@ def register_dynamic_quantized_layer(clso):
 class DynamicQDQLayer(SaveableLayerMixin, keras.layers.Layer):
     """Layer that applies dynamic quantize-dequantize to activations."""
 
-    def __init__(self, name, activation_dtype, dtype="float32", asymmetric=False):
+    def __init__(self, name, activation_dtype, dtype="float32", asymmetric=False, fixed_range=None):
         """Initialize the dynamic QDQ helper layer.
 
         Args:
@@ -76,6 +76,8 @@ class DynamicQDQLayer(SaveableLayerMixin, keras.layers.Layer):
             activation_dtype (jnp.dtype): Activation dtype used for quantization.
             dtype (str): dtype for the layer - see keras.layers.Layer API for details.
             asymmetric (bool): Whether to use asymmetric quantization.
+            fixed_range (Optional[Tuple[float, float]]): If provided, use this (min, max) range
+                instead of computing min/max dynamically per batch.
 
         Returns:
             None: Initializes the layer instance.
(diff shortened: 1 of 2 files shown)


neural_compressor/jax/quantization/layers_dynamic.py
```diff
@@ -158,8 +170,36 @@ def call(self, inputs, mask=None):
...
+        """Apply asymmetric quantization using a pre-computed fixed scale.
+
+        Args:
+            inputs (jnp.ndarray): Input tensor.
+            mask (Optional[jnp.ndarray]): Optional mask tensor.
+
+        Returns:
+            jnp.ndarray: Quantized-dequantized tensor.
+        """
+        return self.call_asymmetric(inputs, self._fixed_a_scale, self._fixed_a_zero_point)
```
bkowalskiINTEL, line 203: I think we could skip creating additional methods for fiexd range by adding this small if-else block here.
```suggestion
            if self._fixed_range is None:
                a_scale, a_zero_point = get_q_params(
                batch_min_max, self.activation_dtype, self.compute_dtype, asymmetric=True
            )
            else:
                a_scale = self._fixed_a_scale
                a_zero_point = self._fixed_a_zero_point
            return self.call_asymmetric(inputs, a_scale, a_zero_point)
        a_scale, _ = get_q_params(batch_min_max, self.activation_dtype, self.compute_dtype, asymmetric=False)
        return self.call_symmetric(inputs, a_scale)

```

neural_compressor/jax/quantization/layers_dynamic.py
```diff
@@ -84,46 +86,56 @@ def __init__(self, name, activation_dtype, dtype="float32", asymmetric=False):
...
         self.adequantfun = get_dequantize_fun(dtype=self.compute_dtype, asymmetric=self._is_asymmetric)
+        if self.fixed_range is not None:
+            fixed_min_max = ops.array(self.fixed_range)
+            a_scale, a_zero_point = get_q_params(
+                fixed_min_max, self.activation_dtype, self.compute_dtype, asymmetric=self._is_asymmetric
+            )
+            self._fixed_a_scale = a_scale
+            if self._is_asymmetric:
+                self._fixed_a_zero_point = a_zero_point
+            self.call = self.call_fixed_asymmetric if self._is_asymmetric else self.call_fixed_symmetric
```
bkowalskiINTEL, line 111: We can remove this if we agree on my comment regarding the additional methods

anko-intel: I rather opt to move all if's to add_variables() and make call function (runtime) as simple as possible. I create separate task for it.

bkowalskiINTEL (approved):
LGTM

PR #2468, opened 2026-05-12
Enable assymetric quantization for all MultiHeadAttention qdq layers

wpietka:
## Type of Change

Improvement

## Description

Currently QDynamicMultiHeadAttention creates 4 separate qdq layers. Two of them are potentially asymmetrical - depending on activation dtype - and two others are always symmetrical. There are two problems here: firstly the symmetrical/asymmetrical policy differs from Static version which allows asymmetrical computation for all qdq layers and secondly dynamic version doesn't need 4 separate qdq layers. Since scale is computed in runtime and not preserved in the layer itself a single qdq layer can be reused for queries, keys and values. Attention qdq stays separate due to fixed range.

Slightly increased dynamic layers accuracy

Vit benchmark has been run with different configurations, and the results show better accuracy with asymmetric layers

## Dependency Change?

No dependency changes


bkowalskiINTEL (approved):
LGTM

PR #2474, opened 2026-05-20
Differentiate between UT/Model Test with custom markers

wpietka:


wpietka:
Complete solution added in https://github.com/intel/neural-compressor/pull/2480

PR #2478, opened 2026-05-26
[Jax] Add support for inplace=False in quantization

wpietka:
## Type of Change

feature

## Description

Adds support for inplace=False in quantize_model
    
Implementation clones original model by calling from_config with custom weights Initializer which copies arrays from original model. This approach functions correctly only for Functional and Sequential models. For models created by subclassing keras.Model this approach doesn't work, since their weights are not stored in original model's config. For them fallback mechanism is used - loop iterates over original model's inputs shapes are uses them to build new model's layers, then another loop assigns variables to new model by copying original values. 

This implementation shows 5% to 10% slow-downs in benchmarks in comparison to inplace=True

quantize_model with parameter inplace=False returns quantized model without any modifications to original model

- test_vit_quantization.py::test_inplace_false
- test_gemma3.py::test_inplace_false

## Dependency Change?

No dependency changes


neural_compressor/jax/quantization/quantize.py
```diff
@@ -13,16 +13,148 @@
...
+        """
+
+        if visited is None:
+            visited = set()
+        key = (id(orig_obj), id(new_obj))
+        if key in visited:
+            return
+        visited.add(key)
+        if hasattr(orig_obj, "proto") and hasattr(new_obj, "set_proto"):
+            if orig_obj.proto is not None and new_obj.proto is None:
```
wpietka: Check if this is necesarry

neural_compressor/jax/quantization/quantize.py
```diff
@@ -13,16 +13,148 @@
...
+                new_obj.set_proto(orig_obj.proto)
+        for attr_name in vars(orig_obj):
+            orig_attr = getattr(orig_obj, attr_name, None)
+            new_attr = getattr(new_obj, attr_name, None)
+            if orig_attr is not None and new_attr is not None and isinstance(orig_attr, keras.layers.Layer):
```
wpietka: Check for None in redundant

neural_compressor/jax/quantization/quantize.py
```diff
@@ -13,16 +13,152 @@
...
+            value = self._value
+            if dtype is not None:
+                value = jnp.array(value, dtype=dtype)
+            if shape is not None and tuple(value.shape) != tuple(shape):
+                value = jnp.reshape(value, shape)
+            return value
+
+        def get_config(self):
+            return {}
+
```
bkowalskiINTEL: I don't really like this class inside a class design. Can we have it as a separate class?

bkowalskiINTEL:
Also I'd avoid creating a whole instance of a class to apply one function. If the clone_model() is essentially stateless, i.e. nothing needs to be kept after this function is called, let's not create a whole class for that.

neural_compressor/jax/quantization/quantize.py
wpietka: No problem with that, but in my mind this class is closely coupled with the parent class and doesn't have much sense outside of it, so for me it looks better nested like that

bkowalskiINTEL: Ok, we can keep it like that

bkowalskiINTEL (approved):
LGTM

bkowalskiINTEL (approved):
Imo that looks better, LGTM

PR #2479, opened 2026-05-27
[JAX] Fix model loading from preset

wpietka:
## Type of Change

Bug fix

## Description

Fixes error during model loading from preset.

When model is loaded using from_preset(), load_own_variables() is being called multiple times, so during second call cleanup is trying to remove already deletes variables. This PR adds a guard to load_own_variables() to return early if the layer was already loaded

Loading quantized presets works correctly

call keras_hub.models.ImageClassifier.from_preset() on quantized model

## Dependency Change?

no dependency changes

### DIFF
--- a/neural_compressor/jax/quantization/saving.py
+++ b/neural_compressor/jax/quantization/saving.py
@@ -160,7 +160,7 @@ class SaveableLayerMixin:
         # Since this function modifies weights and removes variables, this behaviour causes crashes during loading.
         # To prevent that we can check if _is_quantized is None, since this should be initial value and is
         # modified later in this function
-        if self._is_quantized is not None:
+        if hasattr(self, "_is_quantized") and self._is_quantized is not None:
             return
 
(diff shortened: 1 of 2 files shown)


neural_compressor/jax/quantization/saving.py
```diff
@@ -155,6 +155,14 @@ def load_own_variables(self, store):
         Returns:
             None: Loads variables into the layer.
         """
+
+        # In some cases load_own_variables() may be called multiple times for the same layer.
+        # Since this function modifies weights and removes variables, this behaviour causes crashes during loading.
+        # To prevent that we can check if _is_quantized is None, since this should be initial value and is
+        # modified later in this function
+        if hasattr(self, "_is_quantized") and self._is_quantized is not None:
```
anko-intel, line 163: is it possible that _is_quantized not exist? 
If such situation occurs, it seems to be only our fault - so crash will be better

anko-intel (approved):

PR #2480, opened 2026-05-27
[JAX] Tests overhaul with custom markers for smoke tests

wpietka:
## Type of Change

Feature

## Description

Minor tests overhaul. Includes:
- Adding custom marker indicating which tests are mandatory to pass before merging PR
- Ability to mark specific combinations of test parameters as mandatory to pass before merging PR
- Further parametrization for existing tests, increasing feature coverage
- Small cleanups and standardization across existing tests

- Whole test suite will cover all possible use cases of neural_compressor_jax
- CI tests will cover more uses cases than previously
- The average execution time of tests through Github Actions doesn't increase much (about 38 minutes)
- Increased readability and consistency of existing tests

Run CI tests: pytest -m smoke_test test/jax 
Run all tests: pytest test/jax

## Dependency Change?

No dependency changes


test/jax/test_gemma3.py
```diff
@@ -50,42 +50,90 @@ def random_string():
...
+@pytest.mark.parametrize("const_vars", [False, True], ids=["const_vars=False", "const_vars=True"])
+@pytest.mark.parametrize("save_as_preset", [False, True], ids=["save_as_preset=False", "save_as_preset=True"])
+@pytest.mark.parametrize("model_dtype", ["float32", "bfloat16"], ids=["model_dtype=float32", "model_dtype=bfloat16"])
 @pytest.mark.parametrize(
-    "dynamic",
+    "quantization_dtype", ["fp8_e4m3", "fp8_e5m2"], ids=["quantization_dtype=fp8_e4m3", "quantization_dtype=fp8_e5m2"]
+)
+@pytest.mark.CI_test_if(
+    [
+        "dynamic=True",
```
anko-intel: Can we make it just one string, like : 
"quantization_dtype=fp8_e4m3-model_dtype=float32-save_as_preset=False-const_vars=False-dynamic=True"
and later in pytest_collection_modifyitems just compare/find this string
I think it will be more readable . It should allows in easy way marks set of test with just one (shorter than full list of parameters) string.

test/jax/conftest.py
```diff
@@ -1,5 +1,14 @@
 import os
 
+import pytest
+
+
+def pytest_configure(config):
+    config.addinivalue_line("markers", "CI_test: marks test to be run in CI")
```
anko-intel:
```suggestion
    config.addinivalue_line("markers", "PR_test: marks test to be run before Pull Request can be accepted")
```
or maybe pre_commit

anko-intel (changes requested):

test/jax/test_gemma3.py
wpietka: That would make implementation simpler to understand but would limit functionality. Right now you can specify which conditions must be met and you don't have to specify all of them. For example you could specify just "dynamic=True", "const_vars=False" which would result in marking all the tests with dynamic quantization and without const scale and weight. If we compare a single string then we are forced to type all parameters due to fixed ordering

test/jax/conftest.py
```diff
@@ -15,3 +24,14 @@ def pytest_sessionstart(session):
...
+
+
+def pytest_collection_modifyitems(items):
+    for item in items:
+        marker_smoke_test_if = item.get_closest_marker("smoke_test_if")
+        if marker_smoke_test_if is not None:
+            test_parameters = item.nodeid[item.nodeid.rfind("[") + 1 : -1]
+
+            for parameter_set_to_be_marked in marker_smoke_test_if.args:
+                if parameter_set_to_be_marked == test_parameters:
```
anko-intel:
```suggestion
                if parameter_set_to_be_marked in test_parameters:
```
to allow some kind of multiple selections

anko-intel (approved):

PR #2491, opened 2026-06-12
Run only smoke tests in run_3x_jax.sh

wpietka:
## Type of Change

validation

## Description

Run on smoke tests in run_3x_jax.sh

Github actions will trigger only smoke tests


anko-intel:
made by accident

PR #2494, opened 2026-06-16
[JAX] Add per-channel quantization for kernels

wpietka:
## Type of Change

Feature

## Description

- Implements per-channel quantization along specified axis for get_q_params()
- Uses above implementation to add per-channel quantization for dense layers' kernels

Increased quantization accuracy

Ran smoke tests

## Dependency Change?

No changes


neural_compressor/jax/utils/utility.py
```diff
@@ -242,11 +242,19 @@ def float_get_scale(orig_weight):
         if 0 in orig_weight.shape:
             # For empty tensor, return scale as 1.0
             return jnp.array(1.0, dtype=compute_dtype)
-        return (
-            (jnp.max(jnp.abs(orig_weight), keepdims=True) / jnp.finfo(dtype).max.astype(orig_weight.dtype))
-            .reshape((1,))
-            .astype(compute_dtype)
-        )
+        dtype_max = jnp.finfo(dtype).max.astype(orig_weight.dtype).astype(orig_weight.dtype)
+        if axis is None:
```
anko-intel: consider making this if to calculate axis, and shape and have max/min call the same

bkowalskiINTEL:
I think the logic behind choosing axis is wrong here. 
For example in Dense layer, kernel's dimensions are (... , N, M). In my opinion the scale should be calculated over NxM "parts" of the tensor, so for a 3x100x100, we can say that there are 3 100x100 matrices and each of these should have it's separate scale, so the scale dimension would be (3,). In the implementation here (axis=-1), the scale dimension would be (100,) and I believe it's wrong.
In case of, for example (Batch, Channel, Width, Height) we would calculate scales over the channel axis, which is the third from the end - kernel: (10, 3, 100, 100) -> scale: (3,)
So the axis for Dense should be -3.
It's more complicated in case of EinsumDense, but for now let's resolve this first.

neural_compressor/jax/quantization/layers_static.py
```diff
@@ -778,11 +778,12 @@ def call_int8(self, inputs):
 class QStaticConv2d(QStaticConv2DMixin, keras.layers.Conv2D):
     """Statically quantized Conv2D layer."""
 
-    # kernel shape is defined as: kernel_shape = self.kernel_size + (input_channel // self.groups, self.filters,)
-    # so the output axis (called filters here) is the last axis
     @property
     def w_quant_axis(self):
-        return tuple(i for i in range(self.kernel.ndim) if i != self.kernel.ndim - 1)
+        if hasattr(self, "_kernel_quant"):
```
anko-intel: it seems to be not required for static as it make it internally in kernel property

neural_compressor/jax/utils/utility.py
```diff
@@ -242,11 +242,10 @@ def float_get_scale(orig_weight):
...
             # For empty tensor, return scale as 1.0
             return jnp.array(1.0, dtype=compute_dtype)
-        return (
-            (jnp.max(jnp.abs(orig_weight), keepdims=True) / jnp.finfo(dtype).max.astype(orig_weight.dtype))
-            .reshape((1,))
-            .astype(compute_dtype)
-        )
+        dtype_max = jnp.finfo(dtype).max.astype(orig_weight.dtype).astype(orig_weight.dtype)
+        max_val = jnp.max(jnp.abs(orig_weight), axis=axis, keepdims=True)
+        max_val = jnp.where(max_val == 0, jnp.array(1.0, dtype=max_val.dtype), max_val)
```
anko-intel: is it real scenario? where max is 0?

neural_compressor/jax/quantization/layers_static.py
```diff
@@ -760,7 +778,9 @@ def call_int8(self, inputs):
 class QStaticConv2d(QStaticConv2DMixin, keras.layers.Conv2D):
     """Statically quantized Conv2D layer."""
 
-    pass
+    @property
+    def w_quant_axis(self):
+        return tuple(i for i in range(self.kernel.ndim) if i != self.kernel.ndim - 1)
```
bkowalskiINTEL, line 783: I'd prefer keeping the definitons in Mixin class - QStaticConv2DMixin in this case.

neural_compressor/jax/utils/utility.py
wpietka: Probably not, but we should not allow the scale to be 0 in any scenario so it's just additional safeguard, maybe redundant but also harmless

neural_compressor/jax/quantization/layers_static.py
wpietka: For Dense and EinsumDense definitions are one layer above mixin (because mixin is shared between them, and axis logic is different), therefore for Conv2d I also keep it one level above mixin even though is this case there's no objective reason to do so. I'd prefer to keep it that way but I'm not sticking to that

neural_compressor/jax/utils/utility.py
anko-intel: I take a deeper look, and it seems to be wrong. Please look at https://github.com/intel/neural-compressor/pull/2494/changes/BASE..1c21b1a98f9108b58efbce0edd2cf9efd02282c3#diff-f04e0ab357cab95220f848f460e83e49466d63a9c52189f9649c8b718e30b399L541
This change causing that we never notice that something was not calibrated.

anko-intel (changes requested):

neural_compressor/jax/utils/utility.py
wpietka: fixed

anko-intel (approved):

neural_compressor/jax/quantization/layers_static.py
bkowalskiINTEL: Ok, You convinced me :) let's keep it that way

PR #2495, opened 2026-06-16
[JAX] Prepare QDQ for separate use quantization and dequantization

anko-intel:
## Type of Change

feature

## Description

Prepare QDQLayer's implementation to be used also as only quantization or dequantization
Align QDQLayer call_quantized implementation to always call 3 steps:
1. get scale function (get scale and zero if needed, const or variable)
2. quantize or nop
3. dequantize or nop

## Dependency Change?


chensuyue:
This feature is not target v3.9, right?

anko-intel:
> This feature is not target v3.9, right?

Yes, it is not prepared for v3.9

wpietka (approved):

bkowalskiINTEL (approved):
LGTM

PR #2496, opened 2026-06-16
[JAX] Fix reversible embedding

bkowalskiINTEL:
This PR adds missing registrations for ReversibleEmbedding, which caused this layer not to be quantized. 
This also revealed an issue with weights loading for QStaticReversibleEmbedding, which this PR also fixes.


anko-intel (approved):
LGTM

PR #2497, opened 2026-06-17
[JAX] Update requirements

anko-intel:
## Type of Change

others

## Description

Update requirements to supported packets versions

## Expected Behavior & Potential Risk

the expected behavior that triggered by this PR 

## How has this PR been tested?

how to reproduce the test (including hardware information)

## Dependency Change?

any library dependency introduced or removed

PR #2499, opened 2026-06-17
[JAX] Update prompt in gemma tests

wpietka:
## Type of Change

bug fix

## Description

Changes prompt in test_text_prompt gemma test to make it less prone to fail 

No random test fails


anko-intel (changes requested):
Do not merge. For now the issue is resolved by changing jax version (https://github.com/intel/neural-compressor/pull/2497) - version jax0.10.1 give wrong results. It needs further investigation .

anko-intel (approved):
LGTM

PR #2501, opened 2026-06-17
Update CPU test agent pool

chensuyue:
## Type of Change

validation

## Description

Update CPU test agent pool


chensuyue:
Those CI issues are not related to pool update.

PR #2503, opened 2026-06-22
[JAX] Use proper package name for Save/Load functionality 

wpietka:
## Type of Change

Bug fix

## Description

Save/Load functionality tries to read INC version from the package, but expects the name to be "neural_compressor_jax". This PR adds fallback to "neural_compressor"

Save and load works even when INC was installed without INC_JAX_ONLY=1 flag

pytest test/jax/test_save_load.py

## Dependency Change?

No changes

### DIFF
--- a/neural_compressor/jax/quantization/saving.py
+++ b/neural_compressor/jax/quantization/saving.py
@@ -92,6 +92,8 @@ class VersionManager:
                 # If neural_compressor_jax is not found, try neural_compressor
                 if package == "neural_compressor_jax":
                     config["_versions"][package] = importlib_metadata.version("neural_compressor")
+                else:
+                    raise
 
     @classmethod
     def check_versions_mismatch(cls, config):
@@ -116,6 +118,8 @@ class VersionManager:
                 # If neural_compressor_jax is not found, try neural_compressor
                 if package == "neural_compressor_jax":
                     current_version = importlib_metadata.version("neural_compressor")
+                else:
+                    raise
 
             if version_in_config != current_version:
                 logger.warning(
--- a/neural_compressor/jax/quantization/saving.py
+++ b/neural_compressor/jax/quantization/saving.py
@@ -75,6 +75,18 @@ class VersionManager:
     _MODULES = ["neural_compressor_jax", "keras", "keras_hub"]
 
     @classmethod
+    def get_installed_inc_package_name(cls):
+        possible_names = ["neural_compressor", "neural_compressor_jax"]
+        for name in possible_names:
+            try:
+                importlib_metadata.version(name)
+                return name
+            except importlib_metadata.PackageNotFoundError:
+                continue
+
+        raise ModuleNotFoundError(f"Package of given name not found. Tried: {possible_names}")
+
+    @classmethod
     def add_versions(cls, config):
         """Insert package versions into the serialized config.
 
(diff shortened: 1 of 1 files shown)


neural_compressor/jax/quantization/saving.py
```diff
@@ -86,7 +86,14 @@ def add_versions(cls, config):
         """
         config["_versions"] = {}
         for package in cls._MODULES:
-            config["_versions"][package] = importlib_metadata.version(package)
+            try:
+                config["_versions"][package] = importlib_metadata.version(package)
```
anko-intel, line 103: I have doubt if we have to store neural_compressor_jax (instead of always store neural_compressor)

neural_compressor/jax/quantization/saving.py
```diff
@@ -105,7 +112,15 @@ def check_versions_mismatch(cls, config):
             )
             return
         for package, version_in_config in versions.items():
-            current_version = importlib_metadata.version(package)
+            try:
+                current_version = importlib_metadata.version(package)
+            except importlib_metadata.PackageNotFoundError:
+                # If neural_compressor_jax is not found, try neural_compressor
+                if package == "neural_compressor_jax":
```
anko-intel: what if model was stored with "neural_compressor" and now neural_compressor_jax is imported?
I think it should be also supported.

anko-intel (changes requested):

neural_compressor/jax/quantization/saving.py
wpietka: fixed

neural_compressor/jax/quantization/saving.py
```diff
@@ -74,6 +74,18 @@ class VersionManager:
 
     _MODULES = ["neural_compressor_jax", "keras", "keras_hub"]
 
+    @classmethod
+    def get_installed_inc_package_name(cls):
```
anko-intel:
```suggestion
    def get_installed_neural_compressor_package_name(cls):
```

or as the only usage is in place to get a version:

```suggestion
    def get_installed_neural_compressor_version(cls):
```
and return version from line 82

neural_compressor/jax/quantization/saving.py
```diff
@@ -86,7 +98,12 @@ def add_versions(cls, config):
         """
         config["_versions"] = {}
         for package in cls._MODULES:
-            config["_versions"][package] = importlib_metadata.version(package)
+            if "neural_compressor" in package:
+                config["_versions"][package] = importlib_metadata.version(
```
anko-intel: Now here, we always store neural_compressor_jax even if actually whole packet neural_compressor is used.
I am not sure, but maybe it is good idea to store what kind of installation was used when model was quantized - lets discuss it offline

anko-intel (changes requested):

neural_compressor/jax/quantization/saving.py
```diff
@@ -74,6 +74,18 @@ class VersionManager:
 
     _MODULES = ["neural_compressor_jax", "keras", "keras_hub"]
```
anko-intel: After offline discussion I think it will be better to store neural_compressor
```suggestion
    _MODULES = ["neural_compressor", "keras", "keras_hub"]
```

anko-intel (approved):
LGTM

PR #2507, opened 2026-06-26
use safe weights-only loading and add deserialization security validation

changwangss:
## Type of Change

others

## Description

Makes WOQ loading safer by using `weights_only=True` by default, with a fallback for older PyTorch versions. (https://jira.devtools.intel.com/browse/ILITV-4261)
Add deserialization security validation(by adding blacklist) to prevent pickle RCE attacks (https://jira.devtools.intel.com/browse/ILITV-4276)

## Dependency Change?

### DIFF
--- a/neural_compressor/torch/algorithms/weight_only/save_load.py
+++ b/neural_compressor/torch/algorithms/weight_only/save_load.py
@@ -292,15 +292,11 @@ class WOQModelLoader:
         """
         try:
             return torch.load(weight_file_path, weights_only=True)
-        except TypeError:
-            # Backward compatibility for old PyTorch versions that do not support
-            # the `weights_only` argument.
-            logger.warning(
-                "Current PyTorch does not support `weights_only` in torch.load. "
-                "Falling back to legacy loading behavior for %s.",
-                weight_file_path,
-            )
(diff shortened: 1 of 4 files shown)


neural_compressor/torch/algorithms/weight_only/save_load.py
```diff
@@ -284,6 +284,30 @@ def load_inc_format_woq_model(self):
...
+            return torch.load(weight_file_path, weights_only=True)
+        except TypeError:
+            # Backward compatibility for old PyTorch versions that do not support
+            # the `weights_only` argument.
+            logger.warning(
+                "Current PyTorch does not support `weights_only` in torch.load. "
+                "Falling back to legacy loading behavior for %s.",
+                weight_file_path,
+            )
+            return torch.load(weight_file_path)
```
xin3he: Not sure whether this exception is allowed, Others LGTM.

changwangss: In that case, I'll just raise an error and exit here. I'll also suggest using an older version of Neural Compressor.

xin3he (approved):
Nice work

PR #2510, opened 2026-06-29
[JAX] Add possibility to call quantization to and dequantization only

anko-intel:
## Type of Change

feature
## Description

After the change, QDQLayers allow using quantize and dequantize separately. The scale (and calibration) is calculated in call_q (the quantization part), and call_dq only dequantizes using the already-calculated scale from the quantization step.
Also test verifying it is added.


bkowalskiINTEL (approved):
LGTM

wpietka (approved):

PR #2521, opened 2026-07-09
add `swe_verified`, `swebench_pro`, `mcp_atlas` agent evaluations.

changwangss:
## Type of Change

validation

## Description

- SWE-bench Verified: Give an agent a real GitHub issue, ask it to fix the code, and see whether the tests pass.
- SWE-bench Pro: Similar idea, but with harder tasks, larger codebases, and changes that often span multiple files.
- MCP-Atlas: Give an agent a set of MCP tools and see whether it can find the right ones, use them correctly, and complete the task step by step.

## Dependency Change?


benchmark/agent_benchmarks/run_swe_verified.sh
```diff
@@ -0,0 +1,222 @@
...
+Environment:
+	VLLM_API_KEY         API key sent to vLLM (default: EMPTY)
+	VLLM_NO_PROXY        Hosts that bypass HTTP proxies (default: vLLM host)
+	VLLM_WAIT_TIMEOUT    Readiness timeout in seconds (default: 300)
+	VLLM_PID_FILE        vLLM PID file (default: logs/vllm_<PORT>.pid)
+	OPENAI_BASE_URL      Override the OpenAI-compatible API base URL
+EOF
+}
+
+stop_vllm() {
```
chensuyue: This may not work in current workflow, because the `run_xxx.sh` only triggered on CPU node.   
In the test side, I will stop the vllm docker directly. From user side, you may update `start_vllm_serve.sh` to `run_vllm_serve.sh` and provide a stop option.   
I think the current design philosophy is to completely decouple vLLM and the agent; they only need to communicate through a port connection.

benchmark/agent_benchmarks/setup_vllm.sh
```diff
@@ -0,0 +1,76 @@
...
+install_standard_vllm() {
+    log "Installing standard vLLM ${VLLM_VERSION} from PyPI"
+    uv pip install "vllm==${VLLM_VERSION}"
+}
+
+install_deepseek_v4_vllm() {
+    log "Installing DeepSeek-V4 vLLM stack"
+    uv pip install -U pip setuptools_rust setuptools_scm
+    uv pip install -U evalscope "lm_eval[api]" "lm-eval[ruler]" transformers datasets
+    uv pip install git+https://github.com/intel/auto-round.git@main
```
chensuyue: Do we need to install AR for vllm server?

chensuyue (approved):

PR #2526, opened 2026-07-14
fix: use context managers for file I/O to prevent resource leaks

lxcxjxhx:
## Motivation

Several places in the codebase open files without using context managers (`with` statements), which can lead to resource leaks if exceptions occur or if the file handle is not explicitly closed. This is a common Python anti-pattern that can cause issues in long-running processes or when processing many files.

## Changes

- **`neural_compressor/torch/algorithms/layer_wise/utils.py`**: Replaced `json.load(open(...))` with `with open(...) as f: json.load(f)` in `load_tensor_from_shard` and `load_tensor_from_safetensors_shard` to ensure file handles are properly closed after reading.

- **`neural_compressor/tensorflow/utils/utility.py`**: Added `__del__` method to `CaptureOutputToFile` class to ensure the temporary file is closed even when the object is not used as a context manager (i.e., `__exit__` is never called).

- **`neural_compressor/evaluation/lm_eval/accuracy.py`**: Replaced `output_path_file.open("w", encoding="utf-8").write(dumped)` with a `with` statement to ensure the output file is properly closed after writing.

## Testing

These are defensive resource management improvements that do not change functional behavior. The existing test suite should continue to pass without modification.

### DIFF
--- a/neural_compressor/evaluation/lm_eval/accuracy.py
+++ b/neural_compressor/evaluation/lm_eval/accuracy.py
@@ -259,7 +259,8 @@ def cli_evaluate(args) -> None:
                 eval_logger.info(f"Logging to Weights and Biases failed due to {e}")
 
         if args.output_path:
-            output_path_file.open("w", encoding="utf-8").write(dumped)
+            with output_path_file.open("w", encoding="utf-8") as f:
+                f.write(dumped)
 
             if args.log_samples:
                 for task_name, config in results["configs"].items():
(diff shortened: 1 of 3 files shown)


lxcxjxhx:
Hi @chensuyue @xin3he, could you please review this PR? The changes are defensive resource management improvements using context managers for file I/O.

I noticed the Model-Test-3x CI check is failing. This appears to be unrelated to the code changes (which only add context managers for file handling). Could you help investigate or re-run the CI if needed?

Thanks!

chensuyue (approved):

xin3he (approved):

PR #2536, opened 2026-07-27
[JAX] Remove redundant tests

wpietka:
## Type of Change

Tests

## Description

Removes redundant gemma tests cases to shorten execution time of Pytest Other. Additionally marks config related tests as smoke tests since they execute almost immediately

Bigger coverage by smoke tests and quicker execution of non-smoke tests

N/A

## Dependency Change?

No changes

### DIFF
--- a/test/jax/test_composable_config.py
+++ b/test/jax/test_composable_config.py
@@ -11,10 +11,14 @@ The shared ``model`` / ``calibration_data`` / ``test_data`` fixtures (see
 conftest.py) supply a model with named Dense layers ``first``/``second``/``third``.
 """
 
+import pytest
 from jax import numpy as jnp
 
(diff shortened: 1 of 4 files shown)


bkowalskiINTEL (changes requested):
I see that You removed **model_dtype** parameter in some of the tests and set bfloat16 to one and fp32 to the other, which I think is a good idea - we still cover both dtypes but the number of long tests drops. 
I was thinking if it would be possible to still have this information in the test "name" (i.e. quantization_dtype=fp8_e4m3-const_vars=False-dynamic=True-model_dtype=float32) just to have this ifnormation visible in the test results, without looking at the tests code?

bkowalskiINTEL (approved):
LGTM

PR #2538, opened 2026-07-30
[JAX] Add tests for print_model

wpietka:
## Type of Change

Tests

## Description

Adds tests for print_model

Automated tests for print_model

pytest test/jax/test_print_model.py

## Dependency Change?

No changes


test/jax/test_print_model.py
```diff
@@ -0,0 +1,265 @@
...
+    def _validate_name(self, dsc):
+        name = dsc.split()[1]
+        assert self.name == name
+
+    def _validate_a_scale(self, dsc, const_vars):
+        i_beg = dsc.find("a_scale")
+        i_end = dsc.find("]", i_beg)
+        a_scale = dsc[i_beg : i_end + 1]
+
+        assert a_scale.startswith(f"a_scale{'(attr)' if const_vars else ''}=[")
```
anko-intel: please add here and in the other place some message to user what is wrong, for example:
```suggestion
        assert a_scale.startswith(f"a_scale{'(attr)' if const_vars else ''}=["), f"a_scale is missing or incorrect: {a_scale}"
```

anko-intel (changes requested):
check for a_zero_point is missing

test/jax/test_print_model.py
```diff
@@ -0,0 +1,265 @@
...
+                layer_ref=model._flatten_layers()[4],
+            ),
+        ]
+
+
+@pytest.mark.parametrize("dynamic", [False, True], ids=["dynamic=False", "dynamic=True"])
+@pytest.mark.parametrize("const_vars", [False, True], ids=["const_vars=False", "const_vars=True"])
+@pytest.mark.smoke_test_if("const_vars=False-dynamic=False", "const_vars=True-dynamic=True")
+def test_print_sequential_model(dynamic, const_vars):
+    process = subprocess.run(["python", "-c", print_sequential_model_cmd(dynamic, const_vars)], capture_output=True)
```
anko-intel: Can we avoid spanning separate process with string as a script? When user would like to debug the issue it will be difficult to for example make a breakpoint in print_model call.

wpietka: Debugger attaches to the subprocess without a problem, you debug the same as you would in the main process

test/jax/test_print_model.py
wpietka: Done

wpietka:
> check for a_zero_point is missing

Done

anko-intel (approved):
LGTM

PR #2543, opened 2026-08-05
[JAX] Clean prepare() and from_dict()

anko-intel:
## Type of Change

some cleanup

## Description
prepare()  functions is intend to be used only internally so not giving default values could catch potential
internal wrong usage and allows easier find where config parameters are really set.
Simplify configs by moving from_dict()  implementation to base config class (JaxBaseConfig).

### DIFF
--- a/neural_compressor/jax/quantization/config.py
+++ b/neural_compressor/jax/quantization/config.py
@@ -21,7 +21,7 @@ from __future__ import annotations
 import json
 import re
 from collections import OrderedDict
-from typing import Callable, Dict, List, NamedTuple, Optional, Tuple, Union
+from typing import Callable, Dict, List, NamedTuple, Optional, Self, Tuple, Union
 
 import jax.numpy as jnp
 import keras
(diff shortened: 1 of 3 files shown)


wpietka (approved):

neural_compressor/jax/quantization/config.py
```diff
@@ -289,18 +289,38 @@ def get_model_info(self, model) -> List[Tuple[str, str]]:
...
+        """
+        return cls(
+            weight_dtype=config_dict.get("weight_dtype", "fp8_e4m3"),
+            activation_dtype=config_dict.get("activation_dtype", "fp8_e4m3"),
+            const_scale=config_dict.get("const_scale", False),
+            const_weight=config_dict.get("const_weight", False),
+            weight_scale_granularity=config_dict.get("weight_scale_granularity", "per_tensor"),
+            white_list=config_dict.get("white_list", DEFAULT_WHITE_LIST),
+            exclude_list=config_dict.get("exclude_list", None),
+        )
```
bkowalskiINTEL, line 322: I think we could do it like that:
```suggestion
        return cls(**config_dict)
```
This way we also make things easier if the default values for the cls change

anko-intel: It looks well, but this way we lost setting default values. If someone give json with missing parameters, there could be an error (or undefined behavior). It could happens in real live as we extend our configs. So I am opposite to this change

PR #2549, opened 2026-08-07
[JAX] Make dot_product_attention usage configurable

anko-intel:
## Type of Change

feature

## Description

Allows you to control whether to use dot_product_attention from the configuration. 
Using dot_product_attention is disabled by default


neural_compressor/jax/quantization/layers_dynamic.py
```diff
@@ -704,6 +717,16 @@ def post_quantization_cleanup(self):
         """
         pass
 
+    def _use_fused_attention_op(self):
+        if not hasattr(self, "_all_devices_cpu"):
+            import jax
+
+            self._all_devices_cpu = all(d.platform == "cpu" for d in jax.devices())
+
+        if self._all_devices_cpu:
```
wpietka: _all_devices_cpu is a little awkward name. I woudn't know what it means without reading definition. Maybe _only_cpu_devices_are_available would be better

neural_compressor/jax/quantization/layers_static.py
```diff
@@ -1237,6 +1260,8 @@ def prepare(
         const_scale,
         const_weight,
         w_quant_granularity,
+        *,
+        dot_product_attention_enable=False,
```
wpietka: Why it cannot be a positional parameter?

anko-intel: it could be positional. I revert to positional and maybe make separate PR with kwargs for all optional parameters

neural_compressor/jax/quantization/layers_dynamic.py
anko-intel: I look into original _use_fused_attention_op definition again and decide to return to simple implementation without storing it in the class.

bkowalskiINTEL:
*enable_dot_product_attention* sounds better than *dot_product_attention_enable* to me

anko-intel:
> _enable_dot_product_attention_ sounds better than _dot_product_attention_enable_ to me

yes, it sounds better. But I assume naming schema which in my opinion is better in long term: object_action or object_feature. And this a reason why I used dot_product_attention_enable

bkowalskiINTEL:
Ok, let's leave it that way

bkowalskiINTEL (approved):
LGTM

PR #2551, opened 2026-08-17
Fix PT2E OPT export with recent transformers

yiliu30:
## Summary
- keep `return_dict` enabled when loading OPT causal LM models in the PT2E static quant example
- avoid the transformers 5.x export failure where `forward()` still expects `outputs.last_hidden_state`
- retain the existing quantization flow while restoring successful model export

## Testing
- `python -u examples/pytorch/nlp/huggingface_models/language-modeling/quantization/static_quant/pt2e/run_clm_no_trainer.py --model facebook/opt-125m --quantize --output_dir /tmp/qmodel_save_path_2`


chensuyue:
@XuehaoSun please validate.

chensuyue (approved):

PR #2555, opened 2026-08-19
[JAX] Make gemma test more resistant

anko-intel:
## Type of Change

validation

## Description
The Gemma3 testcases fails after quantisation to a less precise data type. Additional verification is required to check whether the models still work, giving them a second chance to provide the correct answer.


bkowalskiINTEL (approved):
LGTM

PR #2580, opened 2026-09-11
Add tests for examples

wpietka:
## Type of Change

test

## Description

Adds test which runs example scripts from examples/jax/keras to ensure they're all working correctly


anko-intel (approved):
LGTM

PR #2584, opened 2026-09-16
[JAX] Sync with Keras v3.15.1

wpietka:
## Type of Change

feature

## Description

* Re-verified the INC JAX quantized layer implementations (neural_compressor/jax/quantization/layers_static.py and layers_dynamic.py) against the latest Keras (v3.15.1) and Keras-hub (v0.31.1) release tags.
* Removed unused QStaticRotaryEmbedding layer


anko-intel (approved):
LGTM

PR #2585, opened 2026-09-16
[JAX] Add AbsMaxObserver

wpietka:
## Type of Change

feature

## Description

Implements new AbsMaxObserver for the JAX static quantization calibration path and wires it into eligible layers.

- MinMaxObserver is used for asymmetric int8 quantization
- AbsMaxObserver is used for every other case

- New unit tests: test/jax/test_observers.py

## Dependency Change?

None


neural_compressor/jax/quantization/layers_static.py
```diff
@@ -153,6 +153,115 @@ def is_calibrated(self):
         return not (jnp.isinf(self.min_val.value).any())
 
 
+class AbsMaxObserver(keras.layers.Layer):
+    """Observer that tracks the running maximum absolute value for calibration.
+
+    Unlike MinMaxObserver, this observer only records the maximum absolute value
```
anko-intel: Don't compare to other, just describe what is for, also remove "small performance gain"

anko-intel (changes requested):
see also:
 max_val = jnp.max(jnp.abs(orig_weight), axis=axis, keepdims=True)
in neural_compressor/jax/utils/utility.py
Maybe we can remove max(abs( there?

neural_compressor/jax/quantization/layers_static.py
wpietka: done

wpietka:
> see also: max_val = jnp.max(jnp.abs(orig_weight), axis=axis, keepdims=True) in neural_compressor/jax/utils/utility.py Maybe we can remove max(abs( there?

get_scale() is used both for activations and kernel. While for activations it's redundant due to AbsMaxObserver, it is still necessary for kernel scale

anko-intel (approved):
LGTM

anko-intel (approved):
LGTM

PR #2593, opened 2026-09-21
Use the non-deprecated CPU autocast API

Shubham-Padkonde:
## Problem

`torch.cpu.amp.autocast()` has been deprecated since PyTorch 2.4. On current PyTorch it warns on every call:

```
$ python -c "import torch, warnings; warnings.simplefilter('always'); torch.cpu.amp.autocast().__enter__()"
FutureWarning: `torch.cpu.amp.autocast(args...)` is deprecated.
Please use `torch.amp.autocast('cpu', args...)` instead.
```

(reproduced on torch 2.14.0+cpu, the version this repository now pins in CI.)

Two IPEX post-quantization helpers still use it:

- `neural_compressor/torch/algorithms/smooth_quant/smooth_quant.py:399`
- `neural_compressor/torch/algorithms/static_quant/static_quant.py:191`

## Fix

Switch both to the documented replacement, `torch.amp.autocast("cpu")`, which takes the device type as its first argument and is otherwise identical — no dtype or other arguments are passed at either site, so behaviour is unchanged.

These are the only two uses of a deprecated `torch.*.amp` entry point in the library; tests and examples use none.

## Verification

Both files compile and `pytest test/common` is unaffected (34 passed). The two call sites themselves are `# pragma: no cover` and need IPEX plus a bf16-capable CPU, so they are not exercised locally.

---

Disclosure: this change was written by Claude Code (Claude Opus 5) working as my agent, at my direction. The output quoted above comes from real runs in my local environment; I am accountable for what is submitted here and will follow up on review feedback.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Shubham-Padkonde:
Corrected the PR base from master to the current main branch. The comparison is now the intended single commit and two changed files, rather than 74 commits and 562 files. The head commit and its existing sign-off are unchanged. The DCO failure references commits that are no longer in this comparison; could a maintainer rerun that check? My rerun request was not available through the API. This follow-up was made with Codex assistance.

Shubham-Padkonde:
A new CI result on the unchanged b5a1388 head shows both `Run PyTorch Model` jobs passing, but `Generate Report` fails in Azure build 76616. The GitHub annotation only says `Bash exited with code 1` (Build log line 15). My attempts to retrieve the Azure timeline/log were reset by the remote host, so I cannot determine the failing report command.

Could a maintainer share the Generate Report error excerpt or rerun that stage if appropriate? I have not changed the model code based only on a report-generation failure.

https://dev.azure.com/lpot-inc/b7121868-d73a-4794-90c1-23135f974d09/_build/results?buildId=76616

Investigation assisted by Codex.

